PR #80, opened 2021-03-10
chore(cli): Uses meroxa.ClientTimeOut and add debug mode via `MEROXA_DEBUG`

raulb:
# Description of change

Depends on https://github.com/meroxa/meroxa-go/pull/22.

https://meroxa.atlassian.net/browse/PLATFORM-77

# Type of change

- [x]  New feature
- [ ]  Bug fix
- [x]  Refactor
- [ ]  Documentation


cmd/add.go
```diff
@@ -88,7 +85,7 @@ var addResourceCmd = &cobra.Command{
 		}
 
 		ctx := context.Background()
-		ctx, cancel := context.WithTimeout(ctx, clientTimeOut)
+		ctx, cancel := context.WithTimeout(ctx, meroxa.ClientTimeOut)
```
owenthereal, line 88: Hmm..I think the client timeout value is more appropriate to stay in the caller of the meroxa-go library and is determined by the CLI. What was the motivation to move this to meroxa-go?

owenthereal (commented):
LGTM except that I think meroxa-go doesn't need to know anything client timeout. This is what the CLI (the caller of the meroxa-go library) should decide

cmd/add.go
raulb: @owenthereal I checked that `meroxa-go` already had [a timeout](https://github.com/meroxa/meroxa-go/pull/22/files#diff-ba9beaa9231e5dd6df64aa81e79fc43a4c78caf08beafdadb628682a443f6775L134) and considering this timeout is related to the client I found it appropriate to reuse it. Would you think otherwise?

raulb: @owenthereal I think you're right. Even though they do coincide these should be separate. Will make that change

owenthereal (approved):

PR #106, opened 2021-04-01
Allow to set pipeline name on connector creation

lovromazgon:
# Description of change

Adds the parameter `--pipeline <pipeline ID>` to the `create connector` command. Also refactors the command to follow the new approach.

Fixes https://meroxa.atlassian.net/browse/PLATFORM-338

# Type of change

- [X]  New feature
- [ ]  Bug fix
- [X]  Refactor
- [ ]  Documentation

# How was this tested?

- [X]  Unit Tests
- [X] Local minikube env with latest platform-api master
- [ ]  Deployed to staging

# Additional references

Depends on https://github.com/meroxa/meroxa-go/pull/28

# Demo
```
$ meroxa create pipeline asdf
Creating Pipeline...
Pipeline asdf successfully created!
$ meroxa list pipeline
 ID   Name    State
==== ====== =========
  3   asdf   healthy
$ meroxa create connector --from store --input stream --pipeline asdf
Creating connector from source store...
Connector store-1-954565 successfully created!
$ meroxa list pipeline
 ID   Name    State
==== ====== =========
  3   asdf   healthy
$ meroxa list connector
 ID      TYPE            NAME                STREAMS             STATE
==== ============= ================ ========================== =========
  4   jdbc-source   store-1-954565   output:                    pending
                                     resource-1-942792-stream

```

I also refactored the `connect` command since it was reusing stuff from the old `create connector` command. It now works the same, except the output is a bit different:
```
$ meroxa connect --from store --to lake
Creating connector from source store...
Connector store-1-984998 successfully created!
Creating connector to destination lake...
Connector lake-2-902992 successfully created!
```


cmd/connect.go
```diff
@@ -17,15 +17,59 @@ limitations under the License.
...
+	destCon, err := destCc.execute(ctx, c)
+	if err != nil {
+		return nil, nil, err
+	}
+	destCc.output(destCon)
+
+	return srcCon, destCon, nil
+}
+
+// ConnectCmd represents the `meroxa add resource` command
```
janelletavares, line 70: is this comment stale?

lovromazgon: Not anymore 😜  Thanks for noticing

cmd/connect.go
```diff
@@ -17,15 +17,59 @@ limitations under the License.
...
+	input, config, source, destination string
+}
+
+func (conn *Connect) setFlags(cmd *cobra.Command) {
+	cmd.Flags().StringVarP(&conn.source, "from", "", "", "source resource name")
+	cmd.MarkFlagRequired("from")
+	cmd.Flags().StringVarP(&conn.destination, "to", "", "", "destination resource name")
+	cmd.MarkFlagRequired("to")
+	cmd.Flags().StringVarP(&conn.config, "config", "c", "", "connector configuration")
+	cmd.Flags().StringVarP(&conn.input, "input", "", "", "command delimeted list of input streams")
```
raulb, line 36: I think this command could include `--pipeline` as well to make that option available since we're using `create connection` under the hood.

cmd/create_connector.go
```diff
@@ -21,162 +21,151 @@ import (
...
-	return v, ok
+func (cc *CreateConnector) setFlags(cmd *cobra.Command) {
+	cmd.Flags().StringVarP(&cc.input, "input", "", "", "command delimited list of input streams")
+	cmd.MarkFlagRequired("input")
+
+	cmd.Flags().StringVarP(&cc.config, "config", "c", "", "connector configuration")
+	cmd.Flags().StringVarP(&cc.metadata, "metadata", "m", "", "connector metadata")
+	cmd.Flags().StringVarP(&cc.source, "from", "", "", "resource name to use as source")
+	cmd.Flags().StringVarP(&cc.destination, "to", "", "", "resource name to use as destination")
+	cmd.Flags().IntVarP(&cc.pipelineID, "pipeline", "", 0, "ID of pipeline to attach connector to") // TODO accept name once we display it in `list connectors` command
```
raulb, line 61: I wonder if we should leave this as `StringVarP` instead and eventually accept both pipeline name and ID once the API supports it.

docs/cmd/meroxa_create_resource.md
```diff
@@ -0,0 +1,44 @@
+## meroxa create resource
```
raulb, line 1: I don't think this should have been generated as part of this PR. I wonder if this was generated by accident while refactoring since I don't see anything in the code that should be causing this now.

raulb (changes requested):
Just a comment regarding a generated document for `meroxa_create_resource`. Otherwise, looks great.

I have a question about `--pipeline` being an `int` rather than `string`. Curious to hear thoughts about that.

cmd/create_connector.go
lovromazgon: As per our call I changed it so that the command accepts only the pipeline name.

raulb (approved):
NIce!

PR #117, opened 2021-04-12
[PLATFORM-429] feat(create:endpoint): Adds feedback

raulb:
# Description of change

It also refactors this command adding tests.

Fixes https://meroxa.atlassian.net/browse/PLATFORM-429

# Type of change

- [x]  New feature
- [ ]  Bug fix
- [x]  Refactor
- [ ]  Documentation

# How was this tested?

- [x]  Unit Tests
- [ ]  Tested in staging

# Demo

**Before this pull-request**

```
$ meroxa create endpoint to-mongo-api --protocol http --stream my-stream
```

**After this pull-request**

```
$ meroxa create endpoint to-mongo-api --protocol http --stream my-stream
Creating endpoint...
Endpoint successfully created!
```


cmd/create.go
```diff
@@ -31,7 +31,7 @@ including connectors.`,
 
 	createCmd.AddCommand((&CreateConnector{}).command())
 	createCmd.AddCommand(CreatePipelineCmd())
```
jayjayjpg, line 33: Unrelated to this change, but one question I had: would it make sense to also expose logs from the `CreatePipelineCmd`? I'm not really sure if this would already automatically by applying the change from line 34 here as well or if this would require more changes - what do you think?

jayjayjpg (approved):
:sparkles:

cmd/create.go
raulb: @jayjayjpg great point. I could simply ship that change (one line), or refactor the command to include more test coverage like this one. I'll do the latter since I think it's worth doing. I'll create a jira ticket and will get it done now. Thanks again for the call out.

raulb: @jayjayjpg I actually noticed we do have logs in that command https://github.com/meroxa/cli/blob/master/cmd/create_pipeline.go#L67 and https://github.com/meroxa/cli/blob/master/cmd/create_pipeline.go#L78. Did you maybe mean to something else?

jayjayjpg: Ah sweet, yes that's exactly what I was looking for - thank you for outlining that!

PR #122, opened 2021-04-14
Generate docs for web

anaptfox:
# Description of change

The goal of this PR is to generate a version of the CLI docs that can be easily integrated to the [web](docs.meroxa.com).

# Type of change

- [ ]  New feature
- [ ]  Bug fix
- [ ]  Refactor
- [x]  Documentation

# Demo

**Before this pull-request**

Only `docs/cmd` were generated. This allows for proper docs in Github.

**After this pull-request**

We now have both `docs/cmd/md` and `docs/cmd/www`. 

There are two to not break docs in GitHub, and compile web-specific docs. The differences between the two:

- use `-` vs `_` (typical for URLs)
- add frontmatter
- uses the `/cli/command` URL (will be how it's represented in the web docs)

PR #124, opened 2021-04-15
refactor: Generation docs

raulb:
# Description of change

These are some suggestions to apply to https://github.com/meroxa/cli/pull/122.

**Changes**

1. Make sure we re-generate docs every time we compile. I recently noticed in a PR that we duplicated some documents by accident for a brief period of time when we generated documentation for a command that we didn't end up implementing. By deleting all of those before, we make sure all documentation applies to an existing and valid command.
2. While I liked that the structure of putting the `web` commands under `/cmd/` I thought that by having a long list of files within that folder would make reaching or discovering the existence of `/web` a bit harder so even though `/web` are also Markdown files, I decided to leave the structure as `/cmd/md` (plain old markdown files), and `/cmd/www` to indicate its metadata you included for frontmatter.
3. The `main()` function started to be a bit long so I moved different steps to their own functions for better readability.

# Type of change

- [ ]  New feature
- [ ]  Bug fix
- [x]  Refactor
- [ ]  Documentation


gen-docs/main.go
```diff
@@ -20,8 +20,7 @@ updatedAt: %s
 title: "%s"
 slug: %s
 url: %s
----
-`
+---`
```
anaptfox, line 23: @raulb I think this new line was needed here. If you look at the front matter it results in something like:

```
---
createdAt: 2021-04-15T10:19:35+02:00
updatedAt: 2021-04-15T10:19:35+02:00
title: "meroxa version"
slug: meroxa-version
url: /cli/meroxa-version/
---## meroxa version
```

raulb: @anaptfox oops, my bad!

anaptfox (approved):

PR #125, opened 2021-04-15
Refactoring 🥷

lovromazgon:
# Description of change

A big refactoring that should make it easier to create new commands and test them.

Highlights of this PR:
* `main.go` was moved from the root of the repository to `cmd/meroxa/main.go`, to follow standard practices of Go projects.
* Added a `log` package where we implement 2 simpler loggers, one is a leveled logger for simple text and one for printing JSON. The reasoning is that if we wanted to we could point each logger to its own destination (e.g. write JSON to a file and text to stdout).
  * By default the output of the JSON logger is discarded, unless the `--json` flag is set to `true`, then it's set to stdout.
  * By default the level of the leveled logger is set to Info, unless the `--json` flag is set to `true`, then it's set to Warn. If the `--debug` flag is set to `true` then the level will be set to Debug, regardless of the `--json` flag.
* The root command is now defined in `cmd/meroxa/root/root.go`, old commands were left intact as much as possible and now live in `cmd/meroxa/root/old`.
* The command `meroxa add resource` was adjusted to the new structure as a proof of concept and now lives in `cmd/meroxa/root/add`. The package structure I propose to follow is to group subcommands in their respective sub packages under `cmd/meroxa/root/[sub command]`.
* Added two utilities under `cmd/meroxa/builder`:
  * `builder.BuildCobraCommand` builds a `*cobra.Command` out of a `builder.Command`. It allows us to pick and choose what we want the cobra command to have based on which interfaces our command implements.
  * `builder.BuildFlags` builds the slice `[]builder.Flag` by parsing struct tags. It means that we can define simple flags in a readable format and with minimal boilerplate code.
* Last but not least we now have the package `cmd/meroxa/global` which contains some things shared by all commands, like the creation of a new meroxa client, the version string, creation of a logger and parsing of the config.

While making these changes I noticed a couple of things:
* The [docs generator](https://github.com/meroxa/cli/pull/125/files#diff-4997324ed5dacfa5731f59bf42aac6ef745c25263c2b64063a2a1086c03803caR52) does not know about the version, since we expect the version to be set through goreleaser. I added a static string there and you'll notice that docs did not change.
* The `--timeout` option is not propagated to the meroxa client, it is always using a [timeout of 5 seconds](https://github.com/meroxa/meroxa-go/blob/48fc0ad6501f77f79042b8006a3449919c135f66/meroxa.go#L19). We should make sure to propagate this option instead of using a context timeout.
* Our Meroxa client accepts an access token which it sets as the bearer token and never refreshes it. I guess that's ok for a CLI application since that is not a long-living process, but using this client on the backend would be a pain, since you would have to handle token refreshes manually. Instead, I propose to change this and use the [oauth2](https://github.com/golang/oauth2) library, which takes care of refreshing the token for us.
* The env variable to change the API URL is called `API_URL` and set in the [meroxa-go](https://github.com/meroxa/meroxa-go/blob/48fc0ad6501f77f79042b8006a3449919c135f66/meroxa.go#L145). I have 2 suggestions there:
  * `API_URL` sounds quite generic, I propose to rather use `MEROXA_API_URL`, this will also match the other env variables we expect that can be supplied in form of a flag (while we're at it, we could add a flag `--api-url` as well).
  * We should move the ENV variable parsing from meroxa-go to the cli. Here we can make assumptions as to how our users should supply this setting, while in meroxa-go I think we shouldn't assume where this setting comes from.
* The config file is saved in `$HOME/meroxa.env` by default. Now, this is completely my personal preference, but I'd rather have it as a hidden file, i.e. `$HOME/.meroxa.env`.

# Type of change

- [ ]  New feature
- [ ]  Bug fix
- [X]  Refactor
- [ ]  Documentation

# How was this tested?

- [X]  Unit Tests
- [ ]  Tested in staging


raulb (approved):
This is fantastic!! I'm sure we'll keep iterating over this, (especially if we consider using  `noun+verb+object` instead), but this is composable enough to make that happen.

For some of the things you noticed I already had items to work on those. Let's sync and make some of these changes.

Thanks for making this big refactor @lovromazgon !

👏

PR #130, opened 2021-04-21
Update meroxa-go, fix observations from refactoring

lovromazgon:
# Description of change

The bulk of the changes happened in package `cmd/meroxa/global`, other line changes came because of additional vendor libs (`golang.org/x/oauth2` brings in lots of stuff).

Summary:
* changes the way we refresh the access token (we rely on meroxa-go now)
* changes the location of the config, now it's stored in the user config dir (relevant explanation: https://github.com/golang/go/issues/29960)
  * for now I also added a fallback that searches for the config file in the old location (`$HOME/meroxa.env`) and copies it to the new location if it's found
* adds the hidden flag `--api-url` (hidden, because it's probably only useful for us internally)
* env variable API_URL is renamed to MEROXA_API_URL
* timeout option is now respected and propagated to the meroxa-go client

Depends on https://github.com/meroxa/meroxa-go/pull/36.

# Type of change

- [ ]  New feature
- [X]  Bug fix
- [X]  Refactor
- [ ]  Documentation

# How was this tested?

- [X]  Unit Tests
- [ ]  Tested in staging


cmd/meroxa/global/global.go
```diff
@@ -17,16 +16,22 @@ var (
...
 	cmd.PersistentFlags().BoolVar(&FlagJSON, "json", false, "output json")
-	cmd.PersistentFlags().StringVar(&flagConfig, "config", "", "config file (default is $HOME/meroxa.env)")
+	cmd.PersistentFlags().StringVar(&flagConfig, "config", "", "config file")
+	cmd.PersistentFlags().StringVar(&flagApiUrl, "api-url", "", "API url")
 	cmd.PersistentFlags().BoolVar(&flagDebug, "debug", false, "display any debugging information")
-	cmd.PersistentFlags().DurationVar(&FlagTimeout, "timeout", time.Second*10, "set the client timeout")
+	cmd.PersistentFlags().DurationVar(&flagTimeout, "timeout", time.Second*10, "set the client timeout")
+
+	if err := cmd.PersistentFlags().MarkHidden("api-url"); err != nil {
+		panic(fmt.Sprintf("could not mark flag as hidden: %v", err))
```
raulb, line 33: 👌

cmd/meroxa/global/client.go
```diff
@@ -17,116 +17,51 @@ limitations under the License.
...
-	accessToken, _, err = refresh(domain, clientID, refreshToken)
-	if err != nil {
-		return "", fmt.Errorf("please login or signup by running 'meroxa login'")
-	}
-	cfg.Set("ACCESS_TOKEN", accessToken)
-	// TODO make sure config is updated on disk
-
-	return accessToken, nil
+	// WithAuthentication needs to be added after WithDumpTransport
+	// to catch requests to auth0
```
raulb, line 48: ❤️

cmd/meroxa/global/config.go
```diff
@@ -23,16 +22,22 @@ func readConfig() (*viper.Viper, error) {
...
 		if err != nil {
-			return nil, fmt.Errorf("could not get home directory: %w", err)
+			return nil, fmt.Errorf("could not create meroxa config directory: %w", err)
 		}
 
-		// Set the base name of the config file, without the file extension.
-		cfg.SetConfigName(defaultConfigFilename)
-		cfg.AddConfigPath(home)
+		cfg.AddConfigPath(configDir)
+		cfg.SetConfigName("config")
```
raulb, line 38: Maybe we should document this in the README.md, that meroxa configuration will be stored elsewhere now.

cmd/meroxa/global/client.go
```diff
@@ -17,116 +17,51 @@ limitations under the License.
...
 	clientID = "2VC9z0ZxtzTcQLDNygeEELV3lYFRZwpb"
-	domain   = "auth.meroxa.io"
 )
 
 func NewClient() (*meroxa.Client, error) {
-	accessToken, err := getAccessToken(Config)
-	if err != nil {
-		return nil, err
+	options := []meroxa.Option{
+		meroxa.WithUserAgent(fmt.Sprintf("Meroxa CLI %s", Version)),
```
raulb, line 34: Where is `Version` being set?

raulb (commented):
Just for the timeout change and taking advantage of a more composable way to instantiate a client, this is worth bringing it. I only have an outstanding question regarding the version string used to specify user-agent.

Great work @lovromazgon 👏

cmd/meroxa/global/client.go
lovromazgon: It's the very first thing that happens in main :) https://github.com/meroxa/cli/blob/bb6889750b2c060d7956c0f8aff08407f5d63305/cmd/meroxa/main.go#L26

We still rely on go-releaser to inject the version for us. We still need to figure out how to get the version when generating documentation. I've got some ideas for that but I didn't act on them yet, I'm not convinced by any particular approach, I'm curious what you think though 🙂

raulb: @lovromazgon what I had in mind is to update the version as part of our tagging process. Leverage GitHub actions if possible. I haven't written down the steps anywhere yet, but that was roughly my idea.

raulb (approved):
Tested this locally and worked as expected. Great work @lovromazgon

PR #140, opened 2021-05-06
[PLATFORM-633] Migrate commands left to builder

raulb:
# Description of change

These commands do conform the new design, but to consolidate an only way to write commands, these are also being migrated to thew new builder. 

While I'm at it, I'll fix some of the things there were in the backlog and will add tests.

Fixes https://meroxa.atlassian.net/browse/PLATFORM-546
Fixes https://meroxa.atlassian.net/browse/PLATFORM-633
Fixes https://meroxa.atlassian.net/browse/PLATFORM-461 (once we're in V2)

## Commands migrated

- [x] api
- [x] login (auth)
- [x] logout  (auth)
- [x] whoami  (auth)
- [x] billing
- [x] open 
- [x] version

# Type of change

- [x]  New feature
- [x]  Bug fix
- [x]  Refactor
- [ ]  Documentation

# How was this tested?

- [x]  Unit Tests
- [ ]  Tested in staging

# Demo

**Before this pull-request**

```
❯ meroxa api get /v1/resources
> get /v1/resources
< 400 Bad Request HTTP/2.0
< Date Thu, 06 May 2021 14:55:03 GMT
< Content-Type text/html
< Content-Length 122
< Server awselb/2.0
<html>
<head><title>400 Bad Request</title></head>
<body>
<center><h1>400 Bad Request</h1></center>
</body>
</html>

❯ .m help
  ...
  login       Login or Sign up to the Meroxa Platform
  logout      Clears local login credentials of the Meroxa Platform
  ...
  whoami      Display the current logged in user

```

**After this pull-request**

```
❯ .m api get /v1/resources
> GET /v1/resources
< 200 OK HTTP/2.0
< Date Thu, 06 May 2021 14:55:17 GMT
< Content-Type application/json; charset=utf-8
< Content-Length 487
< Meroxa-Request-Id cbfc9316-1eba-4fd9-a607-4add7d12efd2
< Vary Origin
[]

❯ MEROXA_V2=only .m help
The Meroxa CLI allows quick and easy access to the Meroxa Data Platform.

Using the CLI you are able to create and manage sophisticated data pipelines
with only a few simple commands. You can get started by listing the supported
resource types:

meroxa resources list --types

Usage:
  meroxa [command]

Available Commands:
  ...
  auth        Authentication commands for Meroxa
  ...

Flags:
      --config string      config file
      --debug              display any debugging information
  -h, --help               help for meroxa
      --json               output json
      --timeout duration   set the client timeout (default 10s)

Use "meroxa [command] --help" for more information about a command.

❯ MEROXA_V2=only .m help auth
Authentication commands for Meroxa

Usage:
  meroxa auth [command]

Available Commands:
  login       Login or Sign up to the Meroxa Platform
  logout      Clears local login credentials of the Meroxa Platform
  whoami      Display the current logged in user


Flags:
  -h, --help   help for auth

Global Flags:
      --config string      config file
      --debug              display any debugging information
      --json               output json
      --timeout duration   set the client timeout (default 10s)

Use "meroxa auth [command] --help" for more information about a command.

```


cmd/meroxa/root/root.go
```diff
@@ -85,28 +85,48 @@ resource types:
...
+// e.g.: `meroxa login` -> `meroxa auth login`.
+func setAliases(cmd *cobra.Command) {
+	aliases := map[string]builder.Command{
+		"login":  &auth.Login{},
+		"logout": &auth.Logout{},
+		"whoami": &auth.WhoAmI{},
+	}
+
+	for _, c := range aliases {
+		cc := builder.BuildCobraCommand(c)
```
raulb, line 128: Note: I could include something like this to make more explicit. I didn't because those three auth commands happen to have the same "Use".

```
for v, c := range aliases {
		cc := builder.BuildCobraCommand(c)
                 cc.Use = v
...
}
```

cmd/meroxa/root/api/api.go
```diff
@@ -1 +1,121 @@
...
+		prettyJSON.Write(b)
+	}
+
+	a.logger.Infof(ctx, "> %s %s\n", a.args.Method, a.args.Path)
+	a.logger.Infof(ctx, "< %s %s\n", resp.Status, resp.Proto)
+	for k, v := range resp.Header {
+		a.logger.Infof(ctx, "< %s %s\n", k, strings.Join(v, " "))
+	}
+
+	a.logger.Info(ctx, prettyJSON.String())
```
lovromazgon, line 118: Should we also add a JSON output to enable the `--json` flag? I imagine it would be quite useful to have in this command. Although the JSON logger right now marshals the output, while in this case we would need to just log the response body as it is, without running it through the JSON marshaler.

cmd/meroxa/root/auth/logout.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+func (l *Logout) Docs() builder.Docs {
+	return builder.Docs{
+		Short: "Clears local login credentials of the Meroxa Platform",
+	}
+}
+
+func (l *Logout) Execute(ctx context.Context) error {
+	// TODO: add confirmation
```
lovromazgon, line 45: Should be pretty easy to add now :)

cmd/meroxa/root/auth/logout.go
```diff
@@ -0,0 +1,54 @@
...
+	global.Config.Set("ACCESS_TOKEN", "")
+	global.Config.Set("REFRESH_TOKEN", "")
+	err := global.Config.WriteConfig()
+	if err != nil {
+		return err
+	}
+	fmt.Println("Successfully logged out.")
```
lovromazgon, line 52: We should rather use the logger.

cmd/meroxa/root/auth/whoami.go
```diff
@@ -0,0 +1,75 @@
...
+package auth
+
+import (
+	"context"
+
+	"github.com/meroxa/cli/cmd/meroxa/builder"
+
+	"github.com/meroxa/cli/log"
+
+	"github.com/meroxa/meroxa-go"
```
lovromazgon, line 26: I heard you like empty lines.. :trollface:

cmd/meroxa/root/billing/billing.go
```diff
@@ -14,32 +14,39 @@ See the License for the specific language governing permissions and
...
 
-	"github.com/meroxa/cli/cmd/meroxa/root/deprecated"
+	"github.com/meroxa/cli/cmd/meroxa/builder"
 
-	"github.com/spf13/cobra"
+	"github.com/meroxa/cli/cmd/meroxa/root/open"
 )
 
-// TODO: Check how to disable parent flags (e.g.: --json)
+type Billing struct{}
```
lovromazgon, line 27: Why do we need a separate command for this if it's doing the same thing? We could just build the same command at a different place (under root instead of under open).

If we really want a separate command we could just do:
```
type Billing struct {
	open.Billing
}
```

And override any methods if needed (doesn't seem like we need to override anything).

cmd/meroxa/root/connectors/connectors.go
```diff
@@ -24,7 +24,8 @@ import (
 type Connectors struct{}
 
 var (
-	_ builder.CommandWithAliases = (*Connectors)(nil)
+	_ builder.CommandWithAliases     = (*Connectors)(nil)
+	_ builder.CommandWithSubCommands = (*Connectors)(nil)
```
lovromazgon, line 28: Nice, while we're at it we're also missing `CommandWithDocs`.
```suggestion
	_ builder.CommandWithSubCommands = (*Connectors)(nil)
	_ builder.CommandWithDocs        = (*Connectors)(nil)
```

cmd/meroxa/root/root.go
lovromazgon: TBH I would use the code in your comment, it doesn't hurt to have it, then we can sleep just a bit easier knowing that if we add more aliases in the future we won't have an unexpected result 🤷

lovromazgon (changes requested):
I've found a bug - when logging in on a fresh machine (no auth.meroxa.io cookie) the auth0 login page opens but after logging in the callback page doesn't open anything, since the server is not running anymore, the CLI process is already stopped at that point.

cmd/meroxa/root/api/api_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+	a := &API{
+		client: client,
+		logger: logger,
+	}
+	a.args.Method = "GET"
+	a.args.Path = "/v1/my-path"
+
+	var httpResponse = &http.Response{
+		StatusCode: 200,
```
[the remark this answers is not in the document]
raulb: @lovromazgon sure, it wouldn't hurt.

cmd/meroxa/root/auth/logout.go
raulb: @lovromazgon oops! missed that one thank you

cmd/meroxa/root/auth/logout.go
raulb: @lovromazgon I actually not sure where this message is coming from. I wouldn't want to require confirmation when logging out. Will remove that comment.

cmd/meroxa/root/billing/billing.go
raulb: @lovromazgon good point.

cmd/meroxa/root/auth/whoami.go
raulb: @lovromazgon blame my IDE! Those are automatically added and since imports are usually collapsed I didn't even notice 😀

cmd/meroxa/root/api/api.go
raulb: @lovromazgon let me know what you think https://github.com/meroxa/cli/pull/140/commits/d8511c9c9a9b46d32a59e63c2c6b9e063c49611d

raulb:
@lovromazgon made some changes. Let me know what you think now. Thank you!

PR #147, opened 2021-06-01
[PLATFORM-734] Fix listing transforms on v1

raulb:
# Description of change

Fixes https://meroxa.atlassian.net/browse/PLATFORM-734

# Type of change

- [ ]  New feature
- [x]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

# Demo

**Before this pull-request**

```
❯ meroxa list transforms
Command "transforms" is deprecated, use `transforms list` instead
... # list resources
```

**After this pull-request**

```
❯ .m list transforms
Command "transforms" is deprecated, use `transforms list` instead
       NAME           TYPE     REQUIRED                                                                         DESCRIPTION                                                                                         PROPERTIES
=================== ========= ========== ========================================================================================================================================================== ===========================================
 InsertField         builtin   false      Add a field using either static data or record metadata                                                                                                    static.field,static.value,timestamp.field
 ReplaceField        builtin   false      Filter or rename fields                                                                                                                                    blacklist,whitelist,renames
 MaskField           builtin   false      Replace field with valid null value for the type (0, empty string, etc) or custom replacement (non-empty string or numeric value only)                     fields,replacement
 ValueToKey          builtin   false      Replace the record key with a new key formed from a subset of fields in the record value                                                                   fields
 HoistField          builtin   false      Wrap the entire event as a single field inside a Struct or a Map                                                                                           field
 ExtractField        builtin   false      Extract a specific field from Struct and Map and include only this field in results                                                                        field
 SetSchemaMetadata   builtin   false      Modify the schema name or version                                                                                                                          schema.name,schema.version
 Filter              builtin   false      This transformation is used to filter records based on a regular expression.                                                                               fields,pattern
 Flatten             builtin   false      Flatten a nested data structure, generating names for each field by concatenating the field names at each level with a configurable delimiter character.   delimiter
                                          Applies to a Struct when a schema is present, or a Map in the case of schemaless data.
```

### DIFF
--- a/cmd/meroxa/root/deprecated/list.go
+++ b/cmd/meroxa/root/deprecated/list.go
@@ -30,6 +30,7 @@ import (
 	"github.com/meroxa/cli/cmd/meroxa/root/endpoints"
 	"github.com/meroxa/cli/cmd/meroxa/root/pipelines"
 	"github.com/meroxa/cli/cmd/meroxa/root/resources"
+	"github.com/meroxa/cli/cmd/meroxa/root/transforms"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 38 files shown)


dianadoherty (approved):

raulb:
@dianadoherty thank you!

PR #148, opened 2021-06-02
[PLATFORM-594] [PLATFORM-595] [PLATFORM-741] Add SSH Tunnel commands

owenthereal:
# Description of change

Add tunnel creation, update, list, and describe. This depends on https://github.com/meroxa/meroxa-go/pull/40. The `meroxa resources describe` layout is changed to accommodate the ssh tunnel display.

Fixes https://meroxa.atlassian.net/browse/PLATFORM-594
Fixes https://meroxa.atlassian.net/browse/PLATFORM-595
Fixes https://meroxa.atlassian.net/browse/PLATFORM-596

# Type of change

- [x]  New feature
- [ ]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

# How was this tested?

- [x]  Unit Tests
- [x]  Tested in staging

# Demo

```
$ m resources create tunnel --type postgres -u DB_URL --ssh-url SSH_TUNNEL_ADDR
$ m resources update tunnel --type postgres -u DB_URL --ssh-url SSH_TUNNEL_ADDR
$ m resource list
  ID      NAME       TYPE                                              URL                                             SSH TUNNELED
====== ========== ========== ======================================================================================== ==============
   11   resource   postgres   postgres://...                                                                               False
 1091   tunnel     postgres   postgres://...                                                                               True
$ m resource describe tunnel
Resource tunnel
================
    ID:   1091
  NAME:   tunnel
  TYPE:   postgres
   URL:   postgres://...
 Status   error

SSH tunnel public key for SSH_TUNNEL_ADDR
=========================================
ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAACAQDP9j0HfeHKDXAoExDpAdplUq0BYDQsIikXkwA3SvEUhYLmliMyZvAXrQQLTlVORdcYPMlFSFILlsPjC58gzwLhfjyi0s+jqNpMaHz4ALu+VazU9zGJmG6TJVF+Vh9Xqvmxe8/v9wucOPyyXTeGkg26BWCfT3YmdCtNrnbbCfl8b83qZg/vhO/rok9+RfpyyVZDwGRXmzHPlSe4h/hp+qb/dqTEh+9KEdO2kKQ7BX5chbcxHuQ9Yu1ul9iKEYi4PBg4F/uzbr7pgXq9CykC+mwG/F0oJeyL2JVhCbKlvXsLIEkTUK2OhbixEfBoSeDDoCidTmd76i4ZM0tvkeHUBbeoIz8SWTF8gVUrxgJCurP+2Wz9IaJYPceemElgztigOFGjvlFvAZG7UT9crB25xTSJ3RAEeZbT8uoPCefkqQ1VGqnMC8jkqHWSOS6KY+en/EBwd+/f77uNlQiBbxWB3iC2NBnshN0lTNfeHV68oezoxfADH06CUpLU7E/3kRKOXVCZwQbuO0NRwmX392IMEmCH32XSps/jcKJ/6bosrfs3iLrj03+qiZXug7tzzn/da36EMmoBrwqW6FDX25BfB7wJX7xUFdT5uXD15Si9hhmSWdmvMYQNVUmHRxZtfqcVfAY99NCT72yCMLLKvEj09a3TG7IzX4Tm1lTRDShdDsgeIQ==
```

# Documentation updated

Not right now until this feature is officially released.

### DIFF
--- a/cmd/meroxa/root/resources/create.go
+++ b/cmd/meroxa/root/resources/create.go
@@ -145,12 +145,12 @@ func (c *Create) Execute(ctx context.Context) error {
 	}
 
 	if tun := res.SSHTunnel; tun == nil {
-		c.logger.Infof(ctx, "%q resource successfully created!", res.Name)
+		c.logger.Infof(ctx, "%q resource is successfully created!", res.Name)
 	} else {
-		c.logger.Infof(ctx, "%q resource created!", res.Name)
+		c.logger.Infof(ctx, "%q resource is successfully created but is pending for validation!", res.Name)
 		c.logger.Info(ctx, "Paste the following public key on your host:")
 		c.logger.Info(ctx, tun.PublicKey)
-		c.logger.Info(ctx, "Meroxa will try to connect to the host for 60 minutes and send email confirmation of a successful resource validation.") //nolint
+		c.logger.Info(ctx, "Meroxa will try to connect to the resource for 60 minutes and send email confirmation after a successful resource validation.") //nolint
 	}
 
 	c.logger.JSON(ctx, res)
--- a/cmd/meroxa/root/resources/create.go
+++ b/cmd/meroxa/root/resources/create.go
@@ -144,7 +144,15 @@ func (c *Create) Execute(ctx context.Context) error {
 		return err
 	}
 
-	c.logger.Infof(ctx, "%q resource successfully created!", res.Name)
+	if tun := res.SSHTunnel; tun == nil {
+		c.logger.Infof(ctx, "%q resource successfully created!", res.Name)
+	} else {
+		c.logger.Infof(ctx, "%q resource created!", res.Name)
+		c.logger.Info(ctx, "Paste the following public key on your host:")
+		c.logger.Info(ctx, tun.PublicKey)
+		c.logger.Info(ctx, "Meroxa will try to connect to the host for 60 minutes and send email confirmation of successful resource validation.")
+	}
+
 	c.logger.JSON(ctx, res)
 
 	return nil
--- a/cmd/meroxa/root/resources/create.go
+++ b/cmd/meroxa/root/resources/create.go
@@ -51,6 +51,7 @@ type Create struct {
 		ClientCert string `long:"client-cert" short:"" usage:"client certificate for authenticating to the resource"`
 		ClientKey  string `long:"client-key"  short:"" usage:"client private key for authenticating to the resource"`
 		SSL        bool   `long:"ssl"         short:"" usage:"use SSL"`
+		SSHURL     string `long:"ssh-url"     short:"" usage:"SSH tunneling address"`
 	}
 }
 
@@ -127,6 +128,12 @@ func (c *Create) Execute(ctx context.Context) error {
 		err := json.Unmarshal([]byte(c.flags.Metadata), &input.Metadata)
 		if err != nil {
 			return fmt.Errorf("could not parse metadata: %w", err)
+		}
+	}
+
+	if sshURL := c.flags.SSHURL; sshURL != "" {
+		input.SSHTunnel = &meroxa.ResourceSSHTunnelInput{
+			Address: sshURL,
 		}
 	}
 
--- a/cmd/meroxa/root/resources/create.go
+++ b/cmd/meroxa/root/resources/create.go
@@ -150,7 +150,7 @@ func (c *Create) Execute(ctx context.Context) error {
 		c.logger.Infof(ctx, "%q resource created!", res.Name)
 		c.logger.Info(ctx, "Paste the following public key on your host:")
 		c.logger.Info(ctx, tun.PublicKey)
-		c.logger.Info(ctx, "Meroxa will try to connect to the host for 60 minutes and send email confirmation of successful resource validation.")
+		c.logger.Info(ctx, "Meroxa will try to connect to the host for 60 minutes and send email confirmation of a successful resource validation.") //nolint
 	}
 
 	c.logger.JSON(ctx, res)
--- a/utils/display.go
+++ b/utils/display.go
@@ -68,6 +68,52 @@ func PrintEndpointsTable(ends []meroxa.Endpoint) {
 	fmt.Println(EndpointsTable(ends))
 }
 
+func header(h string) string {
+	header := h + "\n"
+	for range header {
+		header += "="
+	}
+	header += "\n"
+
+	return header
+}
+
+func ResourceTable(res *meroxa.Resource) string {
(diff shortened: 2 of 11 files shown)


cmd/meroxa/root/resources/create.go
```diff
@@ -51,6 +51,7 @@ type Create struct {
 		ClientCert string `long:"client-cert" short:"" usage:"client certificate for authenticating to the resource"`
 		ClientKey  string `long:"client-key"  short:"" usage:"client private key for authenticating to the resource"`
 		SSL        bool   `long:"ssl"         short:"" usage:"use SSL"`
+		SSHURL     string `long:"ssh-url"     short:"" usage:"SSH tunneling address"`
```
raulb, line 54: Shouldn't this flag be hidden until we make this publicly available?

utils/display.go
```diff
@@ -77,6 +123,7 @@ func ResourcesTable(resources []*meroxa.Resource) string {
 				{Align: simpletable.AlignCenter, Text: "NAME"},
 				{Align: simpletable.AlignCenter, Text: "TYPE"},
 				{Align: simpletable.AlignCenter, Text: "URL"},
+				{Align: simpletable.AlignCenter, Text: "SSH TUNNELED"},
```
raulb, line 126: Is it ok we start showing this column even until we release this feature?

utils/display.go
```diff
@@ -68,6 +68,52 @@ func PrintEndpointsTable(ends []meroxa.Endpoint) {
...
+	for range header {
+		header += "="
+	}
+	header += "\n"
+
+	return header
+}
+
+func ResourceTable(res *meroxa.Resource) string {
+	mainHeader := header(fmt.Sprintf("Resource %s", res.Name))
```
raulb, line 82: I'm not sure I'm in love with this header. It seems kinda redundant if you look at it right next to the command you just executed:

```
❯ .m resource describe resource
Resource resource
==================
    ID:   444
  NAME:   resource
  TYPE:   postgres
   URL:   url
 Status   starting
```

Maybe we should get rid of it.

raulb (commented):
Got some questions regarding the output formatting and wether we should hide certain things such as flag, and `SSH_TUNNELED` column until we release this to everyone.

utils/display.go
```diff
@@ -68,6 +68,52 @@ func PrintEndpointsTable(ends []meroxa.Endpoint) {
...
+		{
+			{Align: simpletable.AlignRight, Text: "TYPE:"},
+			{Text: res.Type},
+		},
+		{
+			{Align: simpletable.AlignRight, Text: "URL:"},
+			{Text: res.URL},
+		},
+		{
+			{Align: simpletable.AlignRight, Text: "Status"},
```
dianadoherty, line 102: Can you rename this to State instead of Status? (for connectors we refer to this field as State (and so does API))

utils/display.go
```diff
@@ -77,6 +123,7 @@ func ResourcesTable(resources []*meroxa.Resource) string {
 				{Align: simpletable.AlignCenter, Text: "NAME"},
 				{Align: simpletable.AlignCenter, Text: "TYPE"},
 				{Align: simpletable.AlignCenter, Text: "URL"},
```
dianadoherty, line 125: Can we have another field here for the State?

owenthereal: Sure, fixed in https://github.com/meroxa/cli/pull/148/commits/df75a278ffa40f0babe2126c5471b22e1151d015

utils/display.go
owenthereal: Sure, fixed in https://github.com/meroxa/cli/pull/148/commits/df75a278ffa40f0babe2126c5471b22e1151d015

utils/display.go
owenthereal: Header was removed in https://github.com/meroxa/cli/pull/148/commits/051d7024e7ebcfe6ac3b9be235657a1f37c3b45d

utils/display.go
raulb: @dianadoherty oh, good point. Thanks @owenthereal for fixing it

cmd/meroxa/root/resources/create.go
owenthereal: Confirmed with products that we are okay with leaving ssh tunnel in the UI for now: https://meroxa.slack.com/archives/C01JL1D0V29/p1623100112056200?thread_ts=1623087396.051400&cid=C01JL1D0V29

owenthereal:
@raulb This is ready to be reviewed again

cmd/meroxa/root/resources/create.go
```diff
@@ -130,14 +131,28 @@ func (c *Create) Execute(ctx context.Context) error {
...
 	res, err := c.client.CreateResource(ctx, &input)
 	if err != nil {
 		return err
 	}
 
-	c.logger.Infof(ctx, "%q resource successfully created!", res.Name)
+	if tun := res.SSHTunnel; tun == nil {
+		c.logger.Infof(ctx, "%q resource successfully created!", res.Name)
+	} else {
+		c.logger.Infof(ctx, "%q resource created!", res.Name)
```
raulb, line 150: This output really should be the same as above (`c.logger.Infof(ctx, "%q resource successfully created!", res.Name)`) and then outside the conditional in consequence.

In other words, remove the `if`/`else` clause, always show that confirmation message once resource is created and then only add the code for the `SSHTunnel`.

raulb (changes requested):
Just a minor tweak on confirmation message on resource creation.

I'm slightly concerned this could be automatically documented in a next build, but it sounds like it _could be ok_.

cmd/meroxa/root/resources/create.go
owenthereal: The message is intentionally omitting "successful". Perhaps there is some misunderstanding of how ssh tunnel works. If a resource has an ssh tunnel, the validation of the resource will be async. So the resource is created but it's not successful until validation goes through until later (customers will need to follow the instruction to put the public key in the bastion etc.). Does it make sense?

owenthereal: Perhaps there is better way to rephrase the message to avoid confusion and make the above more clear.

owenthereal: I took a stab: https://github.com/meroxa/cli/pull/148/commits/7bfc76e4770fe8b1888eb144cd10e80e040b9b10. What do you think now? @raulb

raulb: @owenthereal I think that's perfect, thank you!

raulb (approved):

PR #149, opened 2021-06-08
[PLATFORM-490] Add CLI metrics

raulb:
# Description of change

This pull-request adds the ability to automatically emit events to https://cased.com/.

- Builder commands can implement a custom `Event()` function which would added to what we're currently tracking. This could be useful for updates: https://github.com/meroxa/cli/blob/raul/add-metrics/cmd/meroxa/builder/builder.go#L143-L146. This would be used in https://github.com/meroxa/cli/blob/raul/add-metrics/cmd/meroxa/builder/builder.go#L340-L378.
- If not desired, an event could be build with `CommandWithoutEvent` and in that case nothing would be emitted: https://github.com/meroxa/cli/blob/raul/add-metrics/cmd/meroxa/builder/builder.go#L149-L152.
- Different options could be used to determine how metrics could be published: https://github.com/meroxa/cli/blob/raul/add-metrics/cmd/meroxa/global/metrics_test.go#L186-L245.

Fixes https://meroxa.atlassian.net/browse/PLATFORM-490

# Type of change

- [x]  New feature
- [ ]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

# How was this tested?

- [x]  Unit Tests
- [ ]  Tested in staging

# Demo

**Before this pull-request**

- No CLI metrics.

**After this pull-request**

- All commands with metrics.

# Additional references

The API handler is still being worked on, so for now, to test this is required to set an API key for cased (use `publish_test`). If you need access or want to test it locally, reach out @raulb .


lovromazgon (approved):
LGTM! 👍

PR #158, opened 2021-06-29
add MEROXA_ prefix

terrancej:
# Description of change

*Add `MEROXA_` prefix to access token and refresh token. My particular use case enables a better experience with the terraform provider. However, others may benefit in the future by not overloading the two environment variables. *

# Type of change

- [ ]  New feature
- [ ]  Bug fix
- [X]  Refactor
- [ ]  Documentation

# How was this tested?

- [X]  Unit Tests
- [ ]  Tested in staging

# Demo

**Before this pull-request**

_Include how it looked before_

**After this pull-request**

_Include how it looks now_

# Additional references

*Any additional links (if appropriate)*

# Documentation updated

*Make sure that our [documentation](https://docs.meroxa.com/) is accordingly updated when necessary.*

You can do that by opening a pull-request to our (🔒 private, for now) repository: https://github.com/meroxa/meroxa-docs.

✨ In the future, there will be a GitHub action taking care of these updates automatically. ✨

*Provide PR link:*

### DIFF
--- a/cmd/meroxa/global/client.go
+++ b/cmd/meroxa/global/client.go
@@ -98,8 +98,8 @@ func GetCLIUserInfo() (actor, actorUUID string, err error) {
 }
 
 func GetUserToken() (accessToken, refreshToken string, err error) {
-	accessToken = Config.GetString("ACCESS_TOKEN")
-	refreshToken = Config.GetString("REFRESH_TOKEN")
+	accessToken = Config.GetString("MEROXA_ACCESS_TOKEN")
+	refreshToken = Config.GetString("MEROXA_REFRESH_TOKEN")
 	if accessToken == "" && refreshToken == "" {
 		// we need at least one token for creating an authenticated client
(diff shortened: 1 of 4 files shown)


cmd/meroxa/global/client.go
```diff
@@ -98,8 +98,8 @@ func GetCLIUserInfo() (actor, actorUUID string, err error) {
 }
 
 func GetUserToken() (accessToken, refreshToken string, err error) {
```
raulb, line 100: @terrancej what do you think about adding compatibility with those that still have these without `MEROXA_` prefix? Something similar to what we do in https://github.com/meroxa/cli/blob/master/cmd/meroxa/global/config.go#L77-L79.

raulb (changes requested):
@terrancej This PR sounds reasonable. Maybe we should add still read from those without `MEROXA_` to add compatibility with existing environments.

PR #193, opened 2021-09-27
feat: Update user info every 24h and adds command with feature flag

raulb:
# Description of change

This pull-request adds the ability to create a CLI command that depends on a Meroxa User Feature Flag. These are returned by the API endpoint `/v1/users/me` and when that API endpoint is used in `meroxa whoami` we also make sure we update this information.

To make sure user information is not stale, we'll start fetching this information every 24hours since last time was updated.

To demo the usage of feature flags, this pull-request also adds the `environment` command which could aliased as `env` or `environment`.

Fixes https://github.com/meroxa/product/issues/113

~Depends on https://github.com/meroxa/meroxa-go/pull/53~.

# Type of change

- [x]  New feature
- [ ]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

# How was this tested?

- [ ]  Unit Tests
- [ ]  Tested in staging

# Demo

## Before this pull-request

- We didn't check for feature flags.

## After this pull-request

**Without feature flag enabled**

```bash
$ meroxa env
Error: feature flag "environments" required. reach out to support@meroxa.com for more information
```

**With feature flag enabled**

```bash
$ meroxa env
Welcome to a new world
```

### DIFF
--- a/cmd/meroxa/builder/builder.go
+++ b/cmd/meroxa/builder/builder.go
@@ -38,7 +38,7 @@ import (
 
 type Command interface {
 	// Usage is the one-line usage message.
-	// Recommended syntax is as follow:
+	// Recommended syntax is as follows:
 	//   [ ] identifies an optional argument. Arguments that are not enclosed in brackets are required.
 	//   ... indicates that you can specify multiple values for the previous argument.
 	//   |   indicates mutually exclusive information. You can use the argument to the left of the separator or the
@@ -162,6 +162,11 @@ type CommandWithSubCommands interface {
 	SubCommands() []*cobra.Command
 }
 
+type CommandWithFeatureFlag interface {
+	Command
+	FeatureFlag() string
+}
+
 // BuildCobraCommand takes a Command and builds a *cobra.Command from it. It figures out if the command implements any
 // other CommandWith* interfaces and configures the cobra command accordingly.
 func BuildCobraCommand(c Command) *cobra.Command {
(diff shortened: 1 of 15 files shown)


cmd/meroxa/builder/builder.go
```diff
@@ -586,3 +596,43 @@ func buildCommandWithSubCommands(cmd *cobra.Command, c Command) {
...
+				return err
+			}
+		}
+
+		flagRequired := v.FeatureFlag()
+		userFeatureFlags := global.Config.GetStringSlice(global.UserFeatureFlagsEnv)
+
+		if !hasFeatureFlag(userFeatureFlags, flagRequired) {
+			return fmt.Errorf("your account does not have access to the %q feature. Reach out to support@meroxa.com for more information", flagRequired)
+		}
```
lovromazgon, line 634: Can this be moved outside of the `PreRunE` function? I think we already have all the info at that point, don't we? If we do, then we could already detect if the feature flag is enabled and un-hide the command (`cmd.Hidden = false`).

lovromazgon: Oh no, I see we read the config only after all commands are built ([here](https://github.com/meroxa/cli/blob/ac1587b0a1ff4f11e85bad081340eeeeb15287cf/cmd/meroxa/global/global.go#L58)). We could consider reading the config before even building the commands ([here](https://github.com/meroxa/cli/blob/ac1587b0a1ff4f11e85bad081340eeeeb15287cf/cmd/meroxa/root/root.go#L46)).

dianadoherty (approved):

cmd/meroxa/builder/builder.go
raulb: For context, @lovromazgon and I paired today to explore this option and we both agreed that the effort to make it work is currently not worth it. Maybe in a future iteration we'll revisit this.

PR #194, opened 2021-09-29
fix: Collision with `--global` flags

raulb:
Meroxa CLI used it before to specify the path of a configuration file, while connectors still use it the same as well.

# Description of change

Updates global configuration flag to `meroxa-config` to free up `--config` for connector commands.

# Type of change

- [ ]  New feature
- [x]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

# How was this tested?

- [ ]  Unit Tests
- [ ]  Tested in staging

# Demo

**Before this pull-request**

```
❯ meroxa help | grep config
  config      Manage your Meroxa CLI configuration
      --config string      config file
```

**After this pull-request**

```
❯ .m help | grep config    
  config      Manage your Meroxa CLI configuration
      --meroxa-config string   meroxa configuration file
```

# Additional references

Fixes https://meroxa.zendesk.com/agent/tickets/121

- [ ] Acceptance tests also need to be updated.

# Documentation updated

https://github.com/meroxa/meroxa-docs/commit/a0d441038856553e326af103296fa3562c0fbde5


ahamidi:
If it's used exclusively to point to a config file, then maybe we should make the command --config-file. 

Incidentally I'm pretty sure I accidentally found a work around by using -c for connector configs.

raulb:
@ahamidi re: `-c` you're right! it didn't occur to me 🤦‍♂️.

Regarding `--config-file`, that also works. I wasn't really sure about what name to use, so I'll take that one as a good suggestion. Thank you.

ahamidi:
Actually we may want to be explicit and name it --global-config-file or --cli-config-file or something along those lines so that we free up the ability to pass connector configs as files too.

ahamidi (approved):
LGTM 👍

PR #217, opened 2021-11-04
absorb new meroxa-go interface; small doc changes

janelletavares:
# Description of change

absorb new meroxa-go interface; small doc changes

Fixes <GitHub Issue>

# Type of change

- [ ]  New feature
- [ ]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

# How was this tested?

- [ ]  Unit Tests
- [ ]  Tested in staging

# Demo

**Before this pull-request**

_Include how it looked before_

**After this pull-request**

_Include how it looks now_

# Additional references

*Any additional links (if appropriate)*

# Documentation updated

*Make sure that our [documentation](https://docs.meroxa.com/) is accordingly updated when necessary.*

You can do that by opening a pull-request to our (🔒 private, for now) repository: https://github.com/meroxa/meroxa-docs.

✨ In the future, there will be a GitHub action taking care of these updates automatically. ✨

*Provide PR link:*


cmd/meroxa/root/connectors/describe.go
```diff
@@ -70,7 +70,7 @@ func (d *Describe) Execute(ctx context.Context) error {
 	return nil
```
dianadoherty, line 70: Not sure if this is part of the refactor, but you might wanna modify the Execute fn: https://github.com/meroxa/cli/pull/217/files#diff-36603a939a11613ba22e3770d59fc2369f352994353cd5838259f60272fe772dR61-R69

```
func (d *Describe) Execute(ctx context.Context) error {
	connector, err := d.client.GetConnectorByNameOrID(ctx, d.args.NameOrID)
	if err != nil {
		return err
	}
...
```

cmd/meroxa/root/connectors/logs.go
```diff
@@ -83,7 +83,7 @@ func (l *Logs) Logger(logger log.Logger) {
 	l.logger = logger
```
dianadoherty, line 83: In execute, `GetConnectorLogs` should now accept nameOrID https://github.com/meroxa/meroxa-go/pull/78/files#diff-c10a5335e8fa840c34013ae4363be88f6b8c7ee5573945ced446d6a5f4afe5b5R11

cmd/meroxa/root/connectors/remove.go
```diff
@@ -63,7 +63,7 @@ func (r *Remove) Execute(ctx context.Context) error {
 		return err
 	}
 
-	err = r.client.DeleteConnector(ctx, con.ID)
+	err = r.client.DeleteConnector(ctx, r.args.Name)
```
dianadoherty, line 66:
```suggestion
	err = r.client.DeleteConnector(ctx, r.args.NameOrID)
```

cmd/meroxa/root/connectors/remove.go
```diff
@@ -63,7 +63,7 @@ func (r *Remove) Execute(ctx context.Context) error {
 		return err
 	}
```
dianadoherty, line 64: You also don't need to retrieve the connector name above anymore https://github.com/meroxa/cli/pull/217/files#diff-51d616b524f4dc5671edd4e71797ce6a7cf9c1c23775b1d85c069ab03fd29efcR61

dianadoherty (commented):
In a couple of places I see reference to NameOrID as just name, it might be worth it to clarify that it can be either

cmd/meroxa/root/connectors/describe.go
janelletavares: the Get functions for all entity types are still divided into just id or just name. until we can change the pattern everywhere to one Get function that processes either name or ID, I just left that pattern as is.

cmd/meroxa/root/connectors/remove.go
janelletavares: this link isn't working for me

janelletavares: there were a couple of remove functions that return the whole object... i didn't change that pattern. i don't know if anyone has come to rely on that pattern by now.

raulb:
@janelletavares started reviewing this pull-request, but I think it lacks some context on the motivation behind it. Maybe adding a link to the change made in meroxa-go would be useful. I appreciate we're moving away from having the CLI to define their own mocks, but I noticed some other UX changes (messaging, etc) that would be great to detail.

Let me know if I can help with this.

janelletavares:
> @janelletavares started reviewing this pull-request, but I think it lacks some context on the motivation behind it. Maybe adding a link to the change made in meroxa-go would be useful. I appreciate we're moving away from having the CLI to define their own mocks, but I noticed some other UX changes (messaging, etc) that would be great to detail.
> 
> Let me know if I can help with this.

Here are the related PRs, if that helps.
https://github.com/meroxa/meroxa-go/pull/77
https://github.com/meroxa/meroxa-go/pull/78
https://github.com/meroxa/meroxa-go/pull/79

cmd/meroxa/global/client.go
```diff
@@ -118,7 +118,7 @@ func GetUserToken() (accessToken, refreshToken string, err error) {
 	return accessToken, refreshToken, nil
 }
 
-func NewClient() (*meroxa.Client, error) {
+func NewClient() (meroxa.Client, error) {
```
raulb, line 121: if `meroxa.Client` is not longer a pointer, how come we could reuse it if we haven't returned it?

cmd/meroxa/root/connectors/connect_test.go
```diff
@@ -85,15 +85,15 @@ func TestConnectExecution(t *testing.T) {
 	c.flags.Pipeline = "my-pipeline"
 
 	cSource := utils.GenerateConnector(0, "")
-	cSource.Metadata = map[string]interface{}{"mx:connectorType": "source"}
+	cSource.Type = meroxa.ConnectorTypeSource
```
raulb, line 88: Why making this change from `Metadata` to `Type`?

cmd/meroxa/root/connectors/connect_test.go
```diff
@@ -103,45 +103,44 @@ func TestConnectExecution(t *testing.T) {
...
 				ResourceID: rSource.ID,
 				Configuration: map[string]interface{}{
-					"key":   "value",
-					"input": "my-resource.Table",
-				},
-				Metadata: map[string]interface{}{
-					"mx:connectorType": "source",
+					"key": "value",
 				},
+				Metadata:     map[string]interface{}{},
```
raulb, line 112: How come we're no longer comparing on `Metadata`?

cmd/meroxa/root/connectors/connect_test.go
```diff
@@ -184,7 +183,7 @@ func TestConnectExecutionNoFlags(t *testing.T) {
 
 	err := c.Execute(ctx)
 
-	expected := "requires either a source (--from) or a destination (--to)"
+	expected := "requires pipeline name (--pipeline)"
```
raulb, line 186: I don't think this is right. `meroxa connect ` should still require either `--from` or `--to`.

cmd/meroxa/root/connectors/remove.go
```diff
@@ -58,12 +58,12 @@ func (r *Remove) ValueToConfirm(_ context.Context) (wantInput string) {
 func (r *Remove) Execute(ctx context.Context) error {
 	r.logger.Infof(ctx, "Removing connector %q...", r.args.Name)
 
-	con, err := r.client.GetConnectorByName(ctx, r.args.Name)
+	con, err := r.client.GetConnectorByNameOrID(ctx, r.args.Name)
```
raulb, line 61: we no longer need to make this API request if we can delete passing the name.

cmd/meroxa/root/connectors/remove_test.go
```diff
@@ -73,12 +73,12 @@ func TestRemoveConnectorExecution(t *testing.T) {
 
 	client.
 		EXPECT().
-		GetConnectorByName(ctx, c.Name).
+		GetConnectorByNameOrID(ctx, c.Name).
```
raulb, line 76: This will go away once we remove the API request we no longer need to make.

cmd/meroxa/root/pipelines/create.go
```diff
@@ -48,29 +44,15 @@ type Create struct {
 	args struct {
 		Name string
 	}
-
-	flags struct {
```
raulb, line 52: How come we're removing this?

cmd/meroxa/root/pipelines/update.go
```diff
@@ -44,9 +40,8 @@ type Update struct {
 	}
 
 	flags struct {
-		State    string `long:"state" usage:"new pipeline state (pause | resume | restart)"`
-		Name     string `long:"name" usage:"new pipeline name"`
-		Metadata string `long:"metadata" short:"m" usage:"new pipeline metadata"`
```
raulb, line 49: Same as `pipelines create`, how come we're removing this?

raulb (changes requested):
Thank you @janelletavares for providing those links with changes made in `meroxa-go`. I still have some questions to what seem to be changing publicly documented flags (`metadata`, and some expected behavior when creating connectors via `meroxa connect`).

Let's pair if you think I'm still missing some context.

cmd/meroxa/root/connectors/connect_test.go
janelletavares: The value we put in isn't the totality of the value we get out, so it got messy. there may be a way to adjust for that, i didn't spend a lot of time on it.

cmd/meroxa/root/connectors/connect_test.go
janelletavares: oh, maybe this is because of moving the pipeline flag code up 20 lines in the next block

cmd/meroxa/root/connectors/create.go
```diff
@@ -93,29 +93,25 @@ func (c *Create) CreateConnector(ctx context.Context) (*meroxa.Connector, error)
 		return nil, fmt.Errorf("could not parse metadata: %w", err)
 	}
 
-	// merge in input
-	if c.flags.Input != "" {
-		config["input"] = c.flags.Input
+	if c.flags.Pipeline == "" {
```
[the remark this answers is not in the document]
janelletavares: Checking the Pipeline flag was about 20 lines below, and I just moved it up. I don't remember why I moved it, maybe it doesn't need to move at all. I don't know why it's checked explicitly and not handled by the required notation. Since input is no longer directly stored in the config map and only assigned to the type field on new line 133, I removed the code that checks if it's non-empty and assigns it to the map.

cmd/meroxa/root/connectors/remove.go
janelletavares: we return the whole connector on line 73, that's the only reason that i kept it. it doesn't appear to be documented behavior in the official developer docs.  do you think anyone is relying on that object to be returned? any warnings before we change that behavior?

cmd/meroxa/root/pipelines/create.go
janelletavares: When discussing with Diana, she mentioned it was either unused in the pipeline code or had no use case for being exposed to customers.

cmd/meroxa/root/pipelines/update.go
janelletavares: I forget which, it's either unused or has no use case for being exposed to customers. The only metadata example I can find in our official developer documentation is around creating a postgres resource.

janelletavares (commented):
thanks for the review!

raulb:
@janelletavares Working on the CLI today I realized I'm blocked on some changes made in `meroxa-go`, so I'll need make some of these changes myself to unblock environments' work, unless this pull-request is broken out in different parts. My biggest concern on this pull-request are the user breaking changes that would need to be communicated to customers (removing support on feature flags).

If we could have only one pull-request that could make the CLI work with the latest of meroxa-go would be great, thank you.

cmd/meroxa/global/client.go
raulb: We just talked about this one in a call, so please ignore me.

cmd/meroxa/root/connectors/remove.go
raulb: Actually, never mind. Since `DeleteConnector` doesn't return the connector, we'll still need to fetch it for the json format.

cmd/meroxa/root/connectors/remove_test.go
raulb: See previous message. This is still needed.

janelletavares:
@raulb if the meroxa-go PR and this PR seem good to you, feel free to squash merge, update the go.mod to unblock you.

raulb (approved):
This LGTM now. We'll evaluate removing the `metadata` flag in a separate pull-request following the usual process:

1. Check usage of this flag to determine a deprecation path (or simply removing it if it hasn't been used)
2. Depending on 1, hide flag, set a deprecation date to remove it altogether
3. Remove support in [meroxa-go](https://github.com/meroxa/meroxa-go)

PR #224, opened 2021-12-07
doc: Add documentation on CLI metrics

raulb:
# Description of change

Adds https://github.com/meroxa/cli/blob/document-cased/docs/metrics.md so at least we have a place where we document how CLI metrics are used and configured.

# Type of change

- [ ]  New feature
- [ ]  Bug fix
- [ ]  Refactor
- [x]  Documentation


jayjayjpg (approved):
This reads great! :sparkles:

PR #226, opened 2021-12-08
Tweak environment flags

janelletavares:
# Description of change

The current format for a config flag (and metadata) can be a bit error prone, so switching that to a string array format.

Also, introducing default values for environment create flags.

If we like it, can consider rolling out to other commands.

Fixes <GitHub Issue>

# Type of change

- [ ]  New feature
- [ ]  Bug fix
- [X]  Refactor
- [ ]  Documentation

# How was this tested?

- [X ]  Unit Tests
- [ ]  Tested in staging

# Demo

**Before this pull-request**

```
Flags:
  -c, --config string     environment configuration based on type and provider (e.g.: --config '{"aws_access_key_id":"my_access_key", "aws_secret_access_key":"my_secret_access_key"}')
```

**After this pull-request**

```
Flags:
  -c, --config strings    environment configuration based on type and provider (e.g.: --config aws_access_key_id=my_access_key --config aws_access_secret=my_access_secret)
```

### DIFF
--- a/cmd/meroxa/root/environments/create.go
+++ b/cmd/meroxa/root/environments/create.go
@@ -18,9 +18,9 @@ package environments
 
 import (
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/manifoldco/promptui"
 	"github.com/meroxa/cli/cmd/meroxa/builder"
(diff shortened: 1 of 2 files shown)


cmd/meroxa/root/environments/create.go
```diff
@@ -105,19 +105,25 @@ func (c *Create) setUserValues(e *meroxa.CreateEnvironmentInput) {
 	}
 }
 
+func stringSliceToMap(input []string) map[string]interface{} {
```
raulb, line 108: This should be extracted in some point as part of builder, once we change how config flags are used in other commands, but I think it's fine we leave it here for now.

raulb (changes requested):
@janelletavares changes on `--config` look great. 

I've got some questions on the default value (one of them being that I miss where we make use of it). Do you mind opening a separate PR only making the change of `--config` so we can iterate on each one independently?  Thank you 🙏

raulb (approved):
👍 thank you!

PR #258, opened 2022-02-14
Add `meroxa functions log`

owenthereal:
## Description of change

Add `meroxa functions log`.

Fixes https://github.com/meroxa/cli/issues/244.

## Type of change

- [x]  New feature
- [ ]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

## How was this tested?

- [x]  Unit Tests
- [ ]  Tested in staging

### DIFF
--- a/cmd/meroxa/root/functions/logs.go
+++ b/cmd/meroxa/root/functions/logs.go
@@ -72,7 +72,7 @@ func (l *Logs) Client(client meroxa.Client) {
 
 func (l *Logs) ParseArgs(args []string) error {
 	if len(args) < 1 {
-		return errors.New("requires function name")
+		return errors.New("requires function name or uuid")
 	}
 
 	l.args.NameOrID = args[0]
--- a/cmd/meroxa/root/functions/logs.go
+++ b/cmd/meroxa/root/functions/logs.go
@@ -20,7 +20,7 @@ var (
 )
 
 type functionLogsClient interface {
-	GetFunctionLogs(ctx context.Context, nameOrID string) (*http.Response, error)
+	GetFunctionLogs(ctx context.Context, nameOrUUID string) (*http.Response, error)
 }
 
 type Logs struct {
@@ -43,7 +43,7 @@ func (l *Logs) Docs() builder.Docs {
 }
 
 func (l *Logs) Execute(ctx context.Context) error {
-	resp, err := l.client.GetFunctionLogs(ctx, l.args.NameOrID)
+	resp, err := l.client.GetFunctionLogs(ctx, l.args.NameOrUUID)
 
 	if err != nil {
 		return err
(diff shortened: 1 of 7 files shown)


cmd/meroxa/root/functions/logs.go
```diff
@@ -0,0 +1,80 @@
...
+type functionLogsClient interface {
+	GetFunctionLogs(ctx context.Context, nameOrID string) (*http.Response, error)
+}
+
+type Logs struct {
+	client functionLogsClient
+	logger log.Logger
+
+	args struct {
+		NameOrID string
```
dianadoherty, line 31: Same as meroxa-go, it's best to refer to it as UUID
```suggestion
		NameOrUUID string
```

cmd/meroxa/root/functions/logs.go
```diff
@@ -0,0 +1,80 @@
...
+	l.logger = logger
+}
+
+func (l *Logs) Client(client meroxa.Client) {
+	l.client = client
+}
+
+func (l *Logs) ParseArgs(args []string) error {
+	if len(args) < 1 {
+		return errors.New("requires function name")
```
dianadoherty, line 75: Not sure if it's the standard, but what about including the uuid as well?
```suggestion
		return errors.New("requires function name or uuid")
```

cmd/meroxa/root/functions/logs.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/meroxa/cli/log"
+	"github.com/meroxa/meroxa-go/pkg/meroxa"
+)
+
+var (
+	_ builder.CommandWithDocs    = (*Logs)(nil)
+	_ builder.CommandWithArgs    = (*Logs)(nil)
+	_ builder.CommandWithClient  = (*Logs)(nil)
+	_ builder.CommandWithLogger  = (*Logs)(nil)
+	_ builder.CommandWithExecute = (*Logs)(nil)
```
dianadoherty, line 19: Did you want to make this command hidden? 
```suggestion
	_ builder.CommandWithExecute = (*Logs)(nil)
	_ builder.CommandWithHidden  = (*Repair)(nil)

```

dianadoherty: If it's not supposed to be hidden, then can you run `make build`? It's supposed to also create documentations for each command

dianadoherty (commented):
~~Can you include the example of how logs look like after running this command? It makes it easier to notice if there are any displaying issues :)~~ Never mind, I read your comment 🤦

cmd/meroxa/root/functions/logs.go
owenthereal: The root command is already marked as hidden: https://github.com/meroxa/cli/blob/5590e3e3efedb8ecf1bd7ed27b510a3f14fed964/cmd/meroxa/root/functions/functions.go#L17. So `make build` won't build any docs for its subcommands

dianadoherty (approved):

PR #262, opened 2022-02-25
feat(apps): Apps create command with feature flag

janelletavares:
## Description of change

meroxa apps create NAME

Fixes https://github.com/meroxa/turbine-project/issues/42

## Type of change

- [x]  New feature
- [ ]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

## How was this tested?

- [x]  Unit Tests
- [x]  Tested in minikube

## Demo

Screenshot from 2022-02-25 14-19-02

### DIFF
--- a/cmd/meroxa/root/apps/create.go
+++ b/cmd/meroxa/root/apps/create.go
@@ -78,6 +78,9 @@ func (c *Create) getLang() (string, error) {
 		if err != nil {
 			return lang, err
 		}
+		if c.flags.Lang != "" && c.flags.Lang != lang {
+			fmt.Sprintf("\nIgnoring language flag.\n")
+		}
 		return lang, nil
 	}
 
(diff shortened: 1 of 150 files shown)


cmd/meroxa/root/apps/create.go
```diff
@@ -0,0 +1,104 @@
...
+
+type Create struct {
+	client createApplicationClient
+	logger log.Logger
+
+	args struct {
+		Name string
+	}
+
+	flags struct {
```
raulb, line 46: We don't need a path since we shouldn't be doing anything with it.

raulb: Looks like you mean `language` ?

raulb (changes requested):
Need to change the flag name (it shouldn't be path, but language instead)

raulb (approved):

cmd/meroxa/root/apps/create.go
```diff
@@ -61,15 +65,31 @@ func (c *Create) Flags() []builder.Flag {
...
-		c.args.Name = args[0]
+	if len(args) < 1 {
+		return errors.New("requires an application name")
 	}
+	c.args.Name = args[0]
 	return nil
 }
 
+func (c *Create) getLang() (string, error) {
+	if path := c.flags.Path; path != "" {
```
janelletavares, line 76: what if path is pwd... we still need that to be specified as a flag?

raulb: > what if path is pwd... we still need that to be specified as a flag?

@janelletavares That's a great question. I'm making specifying the path mandatory (notice I'm not reading from pwd) but maybe we could make that change.

> what if the supplied language is different from the app.json language?

Language specified in the app.json currently takes precedence (i.e. if both provided we ignore the currently ignore `--lang`

PR #268, opened 2022-03-04
feat(apps): Absorb turbine changes to create a .gitignore when doing git initialize

janelletavares:
…alize

## Description of change

Fixes https://github.com/meroxa/turbine-project/issues/76

## Type of change

- [ ]  New feature
- [x]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

## How was this tested?

- [ ]  Unit Tests
- [x]  Local manual test
- [ ]  Tested in staging

## Demo

Screenshot from 2022-03-04 11-20-23

### DIFF
--- a/cmd/meroxa/root/apps/init.go
+++ b/cmd/meroxa/root/apps/init.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
 	"os/exec"
 
 	"github.com/meroxa/cli/cmd/meroxa/builder"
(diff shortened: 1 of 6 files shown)


cmd/meroxa/root/apps/init.go
```diff
@@ -64,12 +65,49 @@ func (i *Init) ParseArgs(args []string) error {
 }
 
 func (i *Init) GitInit(ctx context.Context, path string) error {
+	// should this be in turbine?
```
raulb, line 68: @janelletavares I think this could be probably be moved to `turbine-cli` in anticipation to, maybe, eventually move it to its own CLI. Let's not worry about that for now if you think that adds scope.

cmd/meroxa/root/apps/init_test.go
```diff
@@ -85,10 +89,17 @@ func TestGitInit(t *testing.T) {
 
 	for _, tt := range tests {
 		cc := &Init{}
-		err := cc.GitInit(context.Background(), tt.path)
+		cc.Logger(global.NewLogger())
```
raulb, line 92: could we use `log.NewTestLogger()` instead?

raulb (changes requested):
@janelletavares I think this should be part of every turbine-lang where we currently have a folder to populate when doing `git init`. I'll work on what could be look like for `go`, and we could chat later about to see what you think.

raulb (approved):

PR #271, opened 2022-03-08
feat(apps): Associate applications with their entities through pipelines

janelletavares:
## Description of change

Fixes https://github.com/meroxa/turbine-project/issues/75

## Type of change

- [ ]  New feature
- [ ]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

## How was this tested?

- [ ]  Unit Tests
- [ ]  Tested in staging

## Demo

| before | after |
|--------|-------|
|||


cmd/meroxa/root/apps/create.go
```diff
@@ -48,7 +48,7 @@ type Create struct {
 
 	flags struct {
 		Path string `long:"path" usage:"path where app was initialized to read its language from app.json (required unless specifying --lang)"`
-		Lang string `long:"lang" short:"l" usage:"language to use (required unless specifying --path)"`
+		Lang string `long:"language" short:"l" usage:"language to use (required unless specifying --path)"`
```
raulb, line 51: I'm not opposed to this change, but was this something requested or something you thought it was worth updating?

raulb (commented):
I see the changes, and I feel like they don't match both title and description of the PR. What am I missing? 🤔

cmd/meroxa/root/apps/create.go
raulb: @janelletavares what docs are you referring to?

janelletavares: the command docs.... examples, usage.  the example referred to --language but the flag was actually --lang.  other commands use --langage, so for consistency, chose to change the flag instead of the doc.

raulb (approved):

PR #275, opened 2022-03-10
feat(turbine): Warn about uncommitted changes

janelletavares:
## Description of change

Fixes https://github.com/meroxa/turbine-project/issues/70

## Type of change

- [x]  New feature
- [ ]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

## How was this tested?

- [ ]  Unit Tests
- [x]  Tested manually locally
- [ ]  Tested in staging

## Demo
Screenshot from 2022-03-10 11-27-56
Screenshot from 2022-03-10 11-28-10

### DIFF
--- a/cmd/meroxa/root/apps/deploy.go
+++ b/cmd/meroxa/root/apps/deploy.go
@@ -21,6 +21,8 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"os/exec"
+	"strings"
 
 	"github.com/volatiletech/null/v8"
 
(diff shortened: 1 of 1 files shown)


janelletavares:
I commented out most of the deploy command to test this

cmd/meroxa/root/apps/deploy.go
```diff
@@ -155,6 +215,8 @@ func (d *Deploy) Execute(ctx context.Context) error {
 		return err
 	}
 
+	d.gitWarnings(ctx)
```
raulb, line 218: We should interrupt the deploying process if there are changes untracked. Maybe, if `gitWarnings` returned an error, we could do that.

raulb (changes requested):
There is a couple of required changes so we stop the execution of the deployment once we have untracked changes. 

Aside, I'd like to make the suggestion of simplifying this a bit. What I think could be simpler is:

1. Run `cmd := exec.Command("git", "status", "--porcelain=v2")`
2. If there are untracked changes, then run `git status` and show the output to the user including a custom message indicating that need to commit first.
3. If everything is tracked continue. 

What do you think? @janelletavares

cmd/meroxa/root/apps/deploy.go
```diff
@@ -143,6 +145,64 @@ func (d *Deploy) createApplication(ctx context.Context) error {
...
+		return
+	}
+	err = os.Chdir(d.path)
+	if err != nil {
+		return
+	}
+
+	cmd := exec.Command("git", "status", "--porcelain=v2")
+	output, err := cmd.Output()
+	if err == nil {
```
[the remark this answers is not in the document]
janelletavares: so there's a way to deploy without having called `init`?

cmd/meroxa/root/apps/deploy.go
janelletavares: so it's not so much a warning as an error? that does make sense because it is the main branch.

raulb: @janelletavares that's correct.

cmd/meroxa/root/apps/deploy.go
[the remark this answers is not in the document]
raulb: @janelletavares we should consider that option, yes. Not what we'd expect, but I think that's possible.

raulb (approved):
👌

PR #277, opened 2022-03-16
feat(apps): Let users deploy to their own docker hub (and other fixes)

raulb:
- Fixes: https://github.com/meroxa/turbine-project/issues/63
- Includes pipeline uuid which is part of https://github.com/meroxa/turbine-project/issues/79 (still needs to include SHA, which will be part of another iteration)
- Fixes an issue with the regex that determines the number of functions of an application.
- Sets up a nicer path to work on:  https://github.com/meroxa/turbine-project/issues/45
- Fixes lack of error output on `apps init` for JS.

## Type of change

- [x]  New feature
- [x]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

## How was this tested?

- [x]  Unit Tests
- [ ]  Tested in staging

## Demo

See slack.


cmd/meroxa/root/apps/deploy.go
```diff
@@ -105,23 +112,65 @@ func (d *Deploy) getDockerHubAccessTokenEnv() string {
...
+		}
+	}
+	return nil
+}
+
+func (d *Deploy) validateLocalDeploymentConfig() error {
+	// Check if users had an old configuration where DockerHub credentials were set as environment variables
+	err := d.validateDockerHubEnvVars()
+	if err != nil {
+		return err
```
jayjayjpg, line 152: Could we also log out the error here using `logger.Error` onto the console or is this already taken care of by returning the `err` variable?

cmd/meroxa/root/apps/deploy.go
```diff
@@ -200,17 +261,21 @@ func (d *Deploy) Execute(ctx context.Context) error {
...
+		deployOuput, err = d.goDeploy.Deploy(ctx, d.path, d.logger)
 	case "js", JavaScript, NodeJs:
-		err = turbineCLI.DeployJSApp(ctx, d.path, d.logger)
+		err = turbineJS.Deploy(ctx, d.path, d.logger)
 	default:
 		return fmt.Errorf("language %q not supported. %s", d.lang, LanguageNotSupportedError)
 	}
-
 	if err != nil {
 		return err
```
jayjayjpg, line 275: If the deploy should fail with an error, would we need to log the error explicitly here via `logger.Error` or is this already taken care of by returning the `err` object?

jayjayjpg (commented):
This reads great! I left a few more clarifying questions to understand the feature better, but this is already looking good to go :sparkles:

cmd/meroxa/root/apps/deploy.go
raulb: @jayjayjpg It's been taken care of. That function would return the error to then to the Execute function which Cobra will print out as an error.

cmd/meroxa/root/apps/deploy.go
```diff
@@ -44,15 +49,17 @@ type createApplicationClient interface {
 
 type Deploy struct {
 	flags struct {
-		Path string `long:"path" description:"path to the app directory"`
+		Path                 string `long:"path" description:"path to the app directory (default is local directory)"`
+		DockerHubUserName    string `long:"docker-hub-username" description:"DockerHub username to use to build and deploy the app"`
+		DockerHubAccessToken string `long:"docker-hub-access-token" description:"DockerHub access token to use to build and deploy the app"`
```
jayjayjpg, line 54: Yes, this reads great and thank you for the swift update! :sparkles:

cmd/meroxa/root/apps/deploy.go
jayjayjpg: Got it and thank you for the outline, that's super helpful!

cmd/meroxa/root/apps/deploy.go
jayjayjpg: Got it, I wasn't fully aware yet, that an error essentially needs to be passed through to different commands and that sounds good!

jayjayjpg (approved):
This reads great! Thank you for the super swift update and all of the details in the PR description, too! :sparkles:

PR #283, opened 2022-03-21
feat: Get image from platform

raulb:
## Description of change

Fixes https://github.com/meroxa/turbine-project/issues/45

~Depends on https://github.com/meroxa/meroxa-go/pull/106~

## Type of change

- [x]  New feature
- [ ]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

## How was this tested?

- [ ]  Unit Tests
- [ ]  Tested in staging

## Demo

See Slack.


cmd/meroxa/root/apps/deploy.go
```diff
@@ -189,13 +198,194 @@ func (d *Deploy) createApplication(ctx context.Context, pipelineUUID, gitSha str
...
+	sourceBlob := meroxa.SourceBlob{Url: s.GetUrl}
+	buildInput := &meroxa.CreateBuildInput{SourceBlob: sourceBlob}
+
+	build, err := d.client.CreateBuild(ctx, buildInput)
+	if err != nil {
+		return "", err
+	}
+
+	fmt.Print("Getting status for build: ", build.Uuid)
+	for {
```
janelletavares, line 320: seems dangerous to have no timeout, but i guess it's easy enough to shutdown the CLI

raulb: @janelletavares that's right, you could still Ctrl+C if you have to. I could still add a time out though

janelletavares:
i've rebased and updated the dependencies for this branch, as well as addressed lint issues. i feel like Deploy is now too complicated to do much more unit testing in a meaningful way without some potentially convoluted refactoring.  i'd prefer to put developer efforts towards acceptance tests as this point.  what do you think?

raulb:
@janelletavares agreed. Especially since we still need to combine this part to be even more generic so each language does less. Ship it?

cmd/meroxa/root/apps/deploy.go
```diff
@@ -189,13 +201,196 @@ func (d *Deploy) createApplication(ctx context.Context, pipelineUUID, gitSha str
...
+		fmt.Printf(".")
```
janelletavares, line 326: are we using print instead of log because of the newlines here?

cmd/meroxa/turbine_cli/utils.go
```diff
@@ -308,3 +311,58 @@ func RunCmdWithErrorDetection(ctx context.Context, cmd *exec.Cmd, l log.Logger)
...
 	return successMsg, nil
 }
+
+// CreateTarAndZipFile creates a .tar.gz file from `src` on current directory.
+func CreateTarAndZipFile(src string, buf io.Writer) error {
+	// Grab the directory we care about (app's directory)
+	appDir := filepath.Base(src)
+
+	// Change to parent's app directory
+	err := os.Chdir(filepath.Dir(src))
```
janelletavares, line 321: to be consistent, this function could cd back to pwd like some of the other functions

cmd/meroxa/root/apps/deploy.go
```diff
@@ -189,13 +201,189 @@ func (d *Deploy) createApplication(ctx context.Context, pipelineUUID, gitSha str
...
+	fileToWrite, err := os.OpenFile(dFile, os.O_CREATE|os.O_RDWR, os.FileMode(0777)) //nolint:gomnd
+	defer func(fileToWrite *os.File) {
+		err = fileToWrite.Close()
+		if err != nil {
+			panic(err.Error())
+		}
+	}(fileToWrite)
+
+	if err != nil {
+		panic(err)
```
janelletavares, line 231: why are we panicking here?  i can see how it makes sense inside the defer function, but i was surprised to see it here and on 234

raulb: @janelletavares you're right. I should deal with it different. I'll change it

janelletavares (approved):

PR #288, opened 2022-03-25
Include all pre-requisites for go app run in app init

janelletavares:
## Description of change

Fixes https://github.com/meroxa/turbine-project/issues/92

## Type of change

- [x]  New feature
- [ ]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

## How was this tested?

- [ ]  Unit Tests
- [ ]  Tested in staging
- [x] manual

## Demo
Screenshot from 2022-03-28 15-19-15

Screenshot from 2022-03-28 15-17-49

## Documentation updated

updated turbine PR as well https://github.com/meroxa/turbine/pull/19


cmd/meroxa/turbine_cli/utils.go
```diff
@@ -17,17 +17,25 @@ import (
 )
 
 type AppConfig struct {
-	Name     string `json:"name"`
-	Language string `json:"language"`
+	Name        string            `json:"name"`
+	Environment string            `json:"environment"`
+	Pipeline    string            `json:"pipeline"`
```
raulb, line 22: I don't think this should be here.

cmd/meroxa/turbine_cli/utils.go
```diff
@@ -17,17 +17,25 @@ import (
 )
 
 type AppConfig struct {
-	Name     string `json:"name"`
-	Language string `json:"language"`
+	Name        string            `json:"name"`
+	Environment string            `json:"environment"`
```
raulb, line 21: Maybe add it once we have Environment support?

raulb (approved):
Got a question on some fields for `AppConfig`. Otherwise looks good to me.

Glad to see this feature being added. Thank you @janelletavares !

cmd/meroxa/turbine_cli/utils.go
janelletavares: yeah, if it's gone everywhere else already, don't need it here. 
```suggestion
```

cmd/meroxa/turbine_cli/utils.go
janelletavares: i just didn't want to be overwriting app.json in such a way that i would lose anything that was already there

raulb: @janelletavares I understand. I feel like that shouldn't be there either, and only add it when we enable support to environments, but I won't become a blocker just for that. I don't think it's that important so feel free to ignore my suggestion.

PR #292, opened 2022-03-29
Add ci workflow

raulb:
## Description of change

Contributes to https://github.com/meroxa/product/issues/301

(once we have this one running, we should safely remove CircleCI)

## Type of change

- [ ]  New feature
- [ ]  Bug fix
- [x]  Refactor
- [ ]  Documentation


.github/workflows/ci.yml
```diff
@@ -11,3 +11,13 @@ jobs:
...
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: ^1.17
+      - name: Run tests
+        run: make test
```
jayjayjpg, line 23: Will this run the same tests as on CircleCI?

If so, I think this CI step would also need the meroxa machine users credentials to be able to download Go modules correctly

It may look similar to these:
- Github Action with SSH authentication during dependency installation: https://github.com/meroxa/turbine-js/pull/35
- CircleCI workflow with SSH authentication for Go modules: https://github.com/meroxa/cli/pull/290

All of the required credentials are stored in 1Password, but let me know if anything's missing anytime!

jayjayjpg (commented):
This is super neat! I left one more recommendation on setting up credentials, but once this is resolved, this should be good to go - thank you for iterating on this!

.github/workflows/ci.yml
jayjayjpg: Nevermind, I had another look at the CI log, but I cannot actually verify what the actual issue is with the tests failing on GH Actions 🤔

raulb: @jayjayjpg the issue is with the apps init command, which relies on having a GOPATH set up which is where to initialize the go application. This issue could be reproduced locally if you run `make tests` and also having your CLI checked out outside your `GOPATH`.

jayjayjpg: @raulb Got it, thank you for the outline! I'm also curious about the setup you implemented to allow downloads from private repositories correctly -- I'd love to learn more if you happen to have time to pair some time this week or next week!

raulb:
Closing for now to signal we're not actively looking into this. I'll revisit this shortly.

PR #295, opened 2022-03-31
Only create the application if it doesn't already exist with the give…

janelletavares:
…n name

## Description of change
allow this logic to slowly make its way into the turbine libraries

Fixes https://github.com/meroxa/turbine-project/issues/125

## Type of change

- [x]  New feature
- [ ]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

## How was this tested?

- [ ]  Unit Tests
- [x]  Tested in staging

## Demo
Screenshot from 2022-03-31 13-30-48

same logic applied to something easier to test HAH

### DIFF
--- a/cmd/meroxa/root/apps/deploy.go
+++ b/cmd/meroxa/root/apps/deploy.go
@@ -197,7 +197,8 @@ func (d *Deploy) createApplication(ctx context.Context, pipelineUUID, gitSha str
 	if err != nil {
 		if strings.Contains(err.Error(), "already exists") {
 			// Double check that the created application has the expected pipeline.
-			app, err := d.client.GetApplication(ctx, appName)
+			var app *meroxa.Application
+			app, err = d.client.GetApplication(ctx, appName)
 			if err != nil {
 				return err
 			}


ericcheatham (approved):
🎉

raulb (changes requested):
@janelletavares if the app was already created, we should double check before returning that the pipeline associated is the one that's intended, otherwise it'd seemed like it worked, when it shouldn't have.

I was thinking we could fetch for application before creating and then comparing pipeline uuid there (if it exists). This, only until each turbine language implements their own app creation. 

What do you think?

janelletavares:
> @janelletavares if the app was already created, we should double check before returning that the pipeline associated is the one that's intended, otherwise it'd seemed like it worked, when it shouldn't have.
> 
> I was thinking we could fetch for application before creating and then comparing pipeline uuid there (if it exists). This, only until each turbine language implements their own app creation.
> 
> What do you think?

double  checking seems very reasonable.  will whip that up.

janelletavares:
small meroxa-go change was needed https://github.com/meroxa/meroxa-go/pull/108/files

raulb (approved):

PR #298, opened 2022-04-01
Distinguish between regular and extended app describe

janelletavares:
## Description of change

Fixes https://github.com/meroxa/turbine-project/issues/111

## Type of change

- [ ]  New feature
- [ ]  Bug fix
- [x]  Refactor
- [ ]  Documentation

## How was this tested?

- [ ]  Unit Tests
- [x]  Tested in staging

## Demo

Screenshot from 2022-03-31 15-50-59
Screenshot from 2022-03-31 19-14-21


raulb (approved):
Yesterday I was chatting with @jayjayjpg about this one, and I was wondering if this information could be returned by the API directly, so we wouldn't need to fetch for each of the resources. Possibly having even two different endpoints such as:

- `GET /v1/applications/:uuid` (no `--extended`)
- `GET /v1/applications/:uuid?extended` (with `--extended`).

What do you think?

P.S..: This question is not a blocker just an alternative implementation in case the list of connectors / functions / etc... ends up being too big.

janelletavares:
> Yesterday I was chatting with @jayjayjpg about this one, and I was wondering if this information could be returned by the API directly, so we wouldn't need to fetch for each of the resources. Possibly having even two different endpoints such as:
> 
> * `GET /v1/applications/:uuid` (no `--extended`)
> * `GET /v1/applications/:uuid?extended` (with `--extended`).
> 
> What do you think?
> 
> P.S..: This question is not a blocker just an alternative implementation in case the list of connectors / functions / etc... ends up being too big.

the current implementation is quite slow.  doing it all in the API would be faster and nicer.  i could go either way, but at this point, i would favor making time for python work over another implementation of this; i'd be happy to open a follow up ticket to improve performance/efficiency around this. how does that sound?

jayjayjpg:
@janelletavares Yes, fully agree with you and that sounds good to me!

janelletavares:
opened this issue https://github.com/meroxa/turbine-project/issues/129

utils/display.go
```diff
@@ -873,53 +872,92 @@ func AppTable(app *meroxa.Application) string {
...
-			}
-			names = append(names, id)
+	subTable = extendedFunctionsTable(functions)
+	if subTable != "" {
+		output += "\n" + subTable
+	}
+	return output
+}
+
+func extendedResourcesTable(resources []*meroxa.Resource, connectors map[string]*meroxa.Connector) string {
```
janelletavares, line 926: i'm still not sure why, but the typical library used for these tables was giving me lots of issues with nil pointers when trying to format these two subtables, so i just went with a less flexible option

raulb:
> the current implementation is quite slow. doing it all in the API would be faster and nicer. i could go either way, but at this point, i would favor making time for python work over another implementation of this; i'd be happy to open a follow up ticket to improve performance/efficiency around this. how does that sound?

@janelletavares makes sense to me

PR #299, opened 2022-04-02
UX improvements for go app init and path flag

janelletavares:
## Description of change
print the success log after git init
warn and skip go mod init if the given path is not within $GOPATH
update init go app example to include $GOPATH in the --path value

bonus:
make it clearer when the given path for app run or app deploy is not an app dir ... can't find name doesn't help the user fix the error quickly

Fixes https://github.com/meroxa/turbine-project/issues/128

## Type of change

- [ ]  New feature
- [x]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

## How was this tested?

- [x]  Unit Tests
- [ ]  Tested in staging

## Demo
Screenshot from 2022-04-01 18-23-10
Screenshot from 2022-04-01 18-21-25
Screenshot from 2022-04-01 18-20-34

### DIFF
--- a/cmd/meroxa/root/apps/init.go
+++ b/cmd/meroxa/root/apps/init.go
@@ -46,9 +46,10 @@ func (*Init) Docs() builder.Docs {
 	return builder.Docs{
 		Short: "Initialize a Meroxa Data Application",
 		Example: "meroxa apps init my-app --path ~/code --lang js" +
-			"meroxa apps init my-app --lang go # will be initialized in current directory" +
+			"meroxa apps init my-app --lang go # will be initialized in a dir called my-app in the current directory" +
+			"meroxa apps init my-app --lang go --path $GOPATH/src/github.com/my.org" +
 			"meroxa apps init my-app --lang go --skip-mod-init # will not initialize the new go module" +
-			"meroxa apps init my-app --lang go --mod-vendor # will initialize the new go module and download dependendies to the vendor directory",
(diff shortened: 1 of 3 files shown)


cmd/meroxa/root/apps/init.go
```diff
@@ -111,8 +112,14 @@ func (i *Init) Execute(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+
+	err = i.GitInit(ctx, i.path+"/"+name)
```
jayjayjpg, line 116: Neat!

jayjayjpg (commented):
This reads great! I tested this locally and the skipping of the module initialization works as expected. Is it possible though to ensure that dependencies get installed (`go get ./...`) even if module initialization is skipped?

I'm not sure if this should be part of this PR, a follow up or if this can be addressed programmatically at all -- what are your thoughts?

goinitpath

janelletavares:
> This reads great! I tested this locally and the skipping of the module initialization works as expected. Is it also possible to install dependencies (`go get ./...`) even if module initialization is skipped?
> 
> I'm not sure if this should be part of this PR, a follow up or if this can be addressed programmatically at all -- what are your thoughts?
> 
> goinitpath

thanks for testing @jayjayjpg ! i feel like if we skip mod init, we should also skip installing module dependencies.  both are dependent being in GOPATH to work predictably. 

Screenshot from 2022-04-04 08-28-03

jayjayjpg (approved):

PR #307, opened 2022-04-08
fix: character on failed deployments (and add colors)

raulb:
## Description of change

In some environments we were showing the wrong character on failed deployments. This pull-request fixes it and adds color.

## Type of change

- [x]  New feature
- [x]  Bug fix
- [x]  Refactor
- [ ]  Documentation

## Demo

image

### DIFF
--- a/cmd/meroxa/root/apps/deploy.go
+++ b/cmd/meroxa/root/apps/deploy.go
@@ -207,8 +207,8 @@ func (d *Deploy) createApplication(ctx context.Context, pipelineUUID, gitSha str
 				return err
 			}
 			if app.Pipeline.UUID.String != pipelineUUID {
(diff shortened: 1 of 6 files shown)


cmd/meroxa/root/apps/deploy.go
```diff
@@ -424,7 +424,7 @@ func (d *Deploy) buildApp(ctx context.Context) error {
 		d.logger.StopSpinner("\t")
```
janelletavares, line 424: what about these... would we want a red X in these cases?

janelletavares: could be part of an enhancement ticket

janelletavares (approved):

cmd/meroxa/root/apps/deploy.go
raulb: @janelletavares great catch. I think we do. Updating...

PR #308, opened 2022-04-11
chore: Sunset soon to be legacy commands

raulb:
## Description of change

- Fixes https://github.com/meroxa/turbine-project/issues/104
- Fixes https://github.com/meroxa/turbine-project/issues/105

## Type of change

- [ ]  New feature
- [ ]  Bug fix
- [x]  Refactor
- [x]  Documentation

## How was this tested?

- [x]  Unit Tests
- [ ]  Tested in staging
- [ ]  Tested in minikube

## Demo

**Before**

- All commands functioning as expected

**Now**

- All commands functioning as expected, but:

1. No public documentation 
2. Shows the following messages **before** execution (command will still work as before):

### Pipelines

```
❯ .m pipelines
Command "pipelines" is deprecated, we encourage you operate with applications via `apps` instead

❯ .m pipelines create
Command "pipelines create" is deprecated, we encourage you to create applications via `apps deploy` instead.

❯ .m pipelines describe
Command "pipelines describe" is deprecated, we encourage you to describe your application via `apps describe` instead.

❯ .m pipelines ls
Command "pipelines list" is deprecated, we encourage you to list your applications via `apps ls` instead.

❯ .m pipelines remove
Command "pipelines remove" is deprecated, we encourage you to remove your application via `apps remove` instead.

❯ .m pipelines update
Command "pipelines update" is deprecated, we encourage you to operate with applications via `apps` instead.
```

### Connect / Connectors

```
❯ .m connect
Command "connect" is deprecated, we encourage you to operate with your applications via `apps` instead.

❯ .m connectors
Command "connectors" is deprecated, we encourage you to operate with your application via `apps` instead.

❯ .m connectors create
Command "connectors create" is deprecated, we encourage you to create applications via `apps deploy` instead.

❯ .m connectors describe
Command "describe" is deprecated, we encourage you to describe your application via `apps describe` instead.

❯ .m connectors list
Command "connectors list" is deprecated, we encourage you to list your applications via `apps ls` instead.

❯ .m connectors logs
Command "connectors logs" is deprecated, we encourage you to operate with your applications via `apps` instead.

❯ .m connectors remove
Command "connectors remove" is deprecated, we encourage you to remove your application via `apps remove` instead.

❯ .m connectors update
Command "connectors update" is deprecated, we encourage you to operate with applications via `apps` instead.
```

## Additional references

⚠️ do not release until we have a changelog explaining this change ⚠️ 

## Documentation updated

[...]

PR #309, opened 2022-04-11
feat: Make `apps` commands publicly available

raulb:
## Description of change

Fixes https://github.com/meroxa/turbine-project/issues/150

## Type of change

- [ ]  New feature
- [ ]  Bug fix
- [ ]  Refactor
- [x]  Documentation

## Additional references

- Build commands no longer require a feature flag but they remain hidden since they should only interact with those commands in case we recommend it via `apps deploy`.

## Documentation updated

- ⚠️  We'll need to update the documentation in docs.meroxa.com once this pull-request and https://github.com/meroxa/cli/pull/308 are both merged so the generated documentation is accurate ⚠️


cmd/meroxa/root/builds/builds.go
```diff
@@ -46,13 +43,9 @@ func (*Builds) Hidden() bool {
...
 }
 
-func (*Builds) FeatureFlag() (string, error) {
-	return "turbine", fmt.Errorf("no access to the Meroxa Data Application feature")
-}
-
 func (*Builds) Docs() builder.Docs {
 	return builder.Docs{
-		Short: "Manage Process Builds on Meroxa",
+		Short: "Operate with Process Builds on Meroxa",
```
janelletavares, line 48: I think this update is less clear than the original.

raulb: @janelletavares ok. I'll change it back since I don't feel strongly opinionated either way. My thinking was, there's nothing to manage on a build (I cannot create, update or delete), but only describe or logs. Maybe `inspect` would make more sense. I'll change it back to `Manage` and if you agree with `Inspect` I'll change it in another commit / pr.

janelletavares: Inspect makes sense.  View.  Describe.  Any of those are good with me.

PR #335, opened 2022-04-25
fix: apps init (updates to latest turbine-go)

raulb:
## Description of change

Fixes the following error by upgrading to latest turbine-go.

```
$ .m apps run
# order-go-app
./app.go:64:18: too many arguments in call to dest.Write
	have (turbine.Records, string, nil)
	want (turbine.Records, string)
Error: build failed
```

## Type of change

- [ ]  New feature
- [x]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

## Demo

**Before**

Seen error above.

**After**

The following works just as expected:

`$ meroxa apps init;meroxa apps run`

### DIFF
--- a/cmd/meroxa/turbine_cli/utils.go
+++ b/cmd/meroxa/turbine_cli/utils.go
@@ -288,7 +288,7 @@ func GoInit(ctx context.Context, l log.Logger, appPath string, skipInit, vendor 
 		if err != nil {
 			return err
 		}
-		depsLog := "Downloading dependencies "
+		depsLog := "Downloading dependencies"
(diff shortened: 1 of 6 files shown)


cmd/meroxa/turbine_cli/utils.go
```diff
@@ -288,7 +288,7 @@ func GoInit(ctx context.Context, l log.Logger, appPath string, skipInit, vendor
 		if err != nil {
 			return err
 		}
-		depsLog := "Downloading dependencies "
+		depsLog := "Downloading dependencies"
```
janelletavares, line 291: this was intentional

PR #347, opened 2022-05-10
feat(config): allow users to run cli in dev mode via env variables

jayjayjpg:
## Description of change

This allows developers to run the CI in a specific development mode that uses a local copy of turbine-j.

Using the `MEROXA_USE_LOCAL_TURBINE_JS` environment variable in `.envrc` or directly via the commnand line will ensure that the local installation of turbine-js is used (instead of the distributed package). In the future, we could add even more local development-specific settings and make this mode more generic.

This also adds the assertion library `testify` which I saw on @nassor 's recent [testing work for Tricorder](https://github.com/meroxa/tricorder/blob/main/internal/adapters/mysql/extractor_int_test.go#L33).

Part of https://github.com/meroxa/turbine-project/issues/85

## Type of change

- [x]  New feature
- [ ]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

## How was this tested?

- [x]  Unit Tests
- [ ]  Tested in staging
- [ ]  Tested in minikube

## Documentation updated

- [x] Readme


cmd/meroxa/turbine_cli/javascript/build.go
```diff
@@ -13,7 +12,8 @@ import (
 func Build(ctx context.Context, l log.Logger, path string) error {
 	// TODO: Handle the requirement of https://github.com/meroxa/turbine-js.git being installed
 	// cd into the path first
-	cmd := exec.Command("npx", "--yes", "@meroxa/turbine-js@0.1.7", "test", path)
+	isDevModeActive := GetDevModeSetting()
```
jayjayjpg, line 15: Here I'm currently running into the error `GetDevModeSetting` is undefined -- do I need to import this util from the `global` package first?

jayjayjpg: It seems `global.GetDevModeSetting` indeed fixed it

nassor (approved):

raulb:
@jayjayjpg we should changelog about this!

jayjayjpg:
@raulb You're right, let me rewatch your demo real quick and create one!

PR #348, opened 2022-05-10
feat: Create nightly build

raulb:
## Description of change

Fixes https://github.com/meroxa/turbine-project/issues/192

## Type of change

- [x]  New feature
- [ ]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

### DIFF
--- a/.github/workflows/nightly-build.yml
+++ b/.github/workflows/nightly-build.yml
@@ -2,10 +2,9 @@ name: Nightly build
 
 on:
   schedule:
-    # only on weekdays
+    # At 00:00 on every day-of-week from Tuesday through Saturday.
     - cron:  '0 0 * * 2-6'
   workflow_dispatch:
-
 jobs:
   nightly-build:
     name: Push tag for nightly build
--- a/.github/workflows/nightly-build.yml
+++ b/.github/workflows/nightly-build.yml
@@ -1,0 +1,41 @@
+name: Nightly build
+
+on:
+  schedule:
+    # only on weekdays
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly-build.yml
```diff
@@ -0,0 +1,41 @@
+name: Nightly build
+
+on:
+  schedule:
+    # only on weekdays
+    - cron:  '0 0 * * 2-6'
```
hariso, line 6: Technically, this is Tue-Sat, and the reason for that is so we get the changes from Friday sooner rather than later (Sat vs Mon).

hariso (approved):

.github/workflows/nightly-build.yml
```diff
@@ -0,0 +1,41 @@
...
+  workflow_dispatch:
```
jayjayjpg, line 7: I think, because there's no further configuration defined for this event, you could also write:

```suggestion
  workflow_dispatch
```

What do you think?

jayjayjpg (approved):
This reads great, thank you for the super swift iteration on this! :sparkles:

.github/workflows/nightly-build.yml
raulb: @hariso I wondered about that when I noticed it in conduit's repo. I'll update the comment to make it clearer

.github/workflows/nightly-build.yml
raulb: @jayjayjpg I believe even when being empty you still need the `:`.

PR #349, opened 2022-05-10
Provide clarity on how the init command and dashboard interact

janelletavares:
## Description of change

Customers may not immediately understand that the init command doesn't make any API calls and the dashboard won't show the app yet

Fixes https://github.com/meroxa/turbine-project/issues/246

## Type of change

- [ ]  New feature
- [ ]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

## How was this tested?

- [ ]  Unit Tests
- [ ]  Tested in staging
- [ ]  Tested in minikube

## Demo

| before | after |
|--------|-------|
|||

### DIFF
--- a/cmd/meroxa/root/apps/init.go
+++ b/cmd/meroxa/root/apps/init.go
@@ -151,7 +151,8 @@ func (i *Init) Execute(ctx context.Context) error {
 	}
 
 	i.logger.Infof(ctx, "Turbine Data Application successfully initialized!\n"+
-		"You can start interacting with Meroxa in your app located at \"%s/%s\"", i.path, name)
+		"You can start interacting with Meroxa in your app located at \"%s/%s\".\n"+
+		"Your Application will not be visible in the Meroxa Dashboard until after deployment.", i.path, name)
 
 	return nil
 }


cmd/meroxa/root/apps/init.go
```diff
@@ -151,7 +151,8 @@ func (i *Init) Execute(ctx context.Context) error {
 	}
 
 	i.logger.Infof(ctx, "Turbine Data Application successfully initialized!\n"+
-		"You can start interacting with Meroxa in your app located at \"%s/%s\"", i.path, name)
+		"You can start interacting with Meroxa in your app located at \"%s/%s\".\n"+
+		"Your Application will not be visible in the Meroxa Dashboard until after deployment.", i.path, name)
```
raulb, line 155: @janelletavares I thought about including a message here regarding `meroxa apps describe %s --extend` since that's being the most useful command to run after deploying. What do you think?

raulb (approved):
Got a question about including `apps describe --extended`. Aside from that, this should be good!

cmd/meroxa/root/apps/init.go
raulb: @janelletavares oops! my bad. I was thinking about `apps deploy`, you're right!

PR #351, opened 2022-05-10
feat: Add support to python3 as possible option

raulb:
## Description of change

We have noticed some customer attempting to use `python3` as one of the languages. 

Adding "support" for this option is fairly trivial and we should do that since documentation uses `python` and `python3` interchangeably.

## Type of change

- [x]  New feature
- [ ]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

## How was this tested?

- [ ]  Unit Tests
- [ ]  Tested in staging
- [ ]  Tested in minikube

## Demo

**Before**

```
$ meroxa apps init --lang python3
language "python3" not supported. Currently, we support "javascript", "golang", and "python"
```

**Now**

```
$ meroxa apps init --lang python3
Initializing application...
```

### DIFF
--- a/cmd/meroxa/root/apps/deploy.go
+++ b/cmd/meroxa/root/apps/deploy.go
@@ -483,7 +483,7 @@ func (d *Deploy) validateLanguage() error {
 		d.lang = GoLang
 	case "js", JavaScript, NodeJs:
 		d.lang = JavaScript
-	case "py", Python:
+	case "py", Python3, Python:
 		d.lang = Python
 	default:
 		return fmt.Errorf("language %q not supported. %s", d.lang, LanguageNotSupportedError)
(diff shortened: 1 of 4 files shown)


ericcheatham (approved):
❤️

cmd/meroxa/root/apps/deploy.go
```diff
@@ -483,7 +483,7 @@ func (d *Deploy) validateLanguage() error {
 		d.lang = GoLang
 	case "js", JavaScript, NodeJs:
 		d.lang = JavaScript
-	case "py", Python:
+	case "py", Python3, Python:
```
janelletavares, line 486: i think there's one more case in deploy....

raulb: @janelletavares that should be normalized by http://github.com/meroxa/cli/blob/ba431e0c47de0dea6ea7d4b43b0258de8e224a67/cmd/meroxa/root/apps/deploy.go#L480-L492 previously!

PR #352, opened 2022-05-11
Resource availability

janelletavares:
## Description of change

Fixes https://github.com/meroxa/turbine-project/issues/85

### Blockers

- [x] merge of https://github.com/meroxa/turbine-go/pull/53
- [x] bump to latest version of turbine-go
- [x] bump to latest version of turbine-js: https://github.com/meroxa/turbine-js/pull/94
- [x] `make build` to auto-generate docs

This is listing resources correctly for Turbine apps written in

- [x] Python
- [x] Go
- [x] JS

## Type of change

- [x]  New feature
- [ ]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

## How was this tested?

- [x]  Unit Tests
- [x]  Tested in staging
- [ ]  Tested in minikube
- [x]  Tested manually

## Demo

avaresource

### DIFF
--- a/cmd/meroxa/root/apps/deploy.go
+++ b/cmd/meroxa/root/apps/deploy.go
@@ -578,7 +578,7 @@ func (d *Deploy) checkResourceAvailability(ctx context.Context) error {
 
 	if errStr != "" {
 		errStr += ";\n\n ⚠️  Run 'meroxa resources list' to verify that the resource names " +
-			"defined in your Turbine app are identical to the resources you have created on the Meroxa platform before deploying again"
+			"defined in your Turbine app are identical to the resources you have created on the Meroxa Platform before deploying again"
 		d.logger.StopSpinnerWithStatus("Resource availability check failed", log.Failed)
 		return fmt.Errorf("%s", errStr)
 	}
--- a/cmd/meroxa/root/apps/deploy.go
+++ b/cmd/meroxa/root/apps/deploy.go
@@ -583,7 +583,7 @@ func (d *Deploy) checkResourceAvailability(ctx context.Context) error {
 		return fmt.Errorf("%s", errStr)
 	}
 
-	d.logger.StopSpinnerWithStatus("Found all required resources!", log.Successful)
+	d.logger.StopSpinnerWithStatus("Can access to your Turbine resources", log.Successful)
 	return nil
 }
 
--- a/cmd/meroxa/root/apps/deploy.go
+++ b/cmd/meroxa/root/apps/deploy.go
@@ -572,7 +572,7 @@ func (d *Deploy) checkResourceAvailability(ctx context.Context) error {
 		}
 	}
 	if errStr != "" {
-		return fmt.Errorf(errStr)
+		return fmt.Errorf("%s", errStr)
 	}
 	return nil
 }
--- a/cmd/meroxa/root/apps/deploy.go
+++ b/cmd/meroxa/root/apps/deploy.go
@@ -531,11 +531,11 @@ func (d *Deploy) checkResourceAvailability(ctx context.Context) error {
 	var resourceNames []string
 	var err error
 	switch d.lang {
-	case "go", GoLang:
+	case GoLang:
 		resourceNames, err = turbineGo.GetResourceNames(ctx, d.logger, d.path, d.appName)
-	case "js", JavaScript, NodeJs:
+	case JavaScript, NodeJs:
 		resourceNames, err = turbineJS.GetResourceNames(ctx, d.logger, d.path, d.appName)
-	case "py", Python:
+	case Python:
 		resourceNames, err = turbinePY.GetResourceNames(ctx, d.logger, d.path, d.appName)
 	default:
 		return fmt.Errorf("language %q not supported. %s", d.lang, LanguageNotSupportedError)
--- a/cmd/meroxa/root/apps/deploy.go
+++ b/cmd/meroxa/root/apps/deploy.go
@@ -525,40 +525,15 @@ func (d *Deploy) validate(ctx context.Context) error {
 	return turbineCLI.ValidateBranch(ctx, d.logger, d.path)
 }
 
-func (d *Deploy) checkResourceAvailability(ctx context.Context) error {
-	d.logger.Infof(ctx, "Checking resource availability for application %q (%s) before deployment...", d.appName, d.lang)
-
-	var resourceNames []string
-	var err error
-	switch d.lang {
-	case GoLang:
-		resourceNames, err = turbineGo.GetResourceNames(ctx, d.logger, d.path, d.appName)
-	case JavaScript, NodeJs:
-		resourceNames, err = turbineJS.GetResourceNames(ctx, d.logger, d.path, d.appName)
-	case Python:
-		resourceNames, err = turbinePY.GetResourceNames(ctx, d.logger, d.path, d.appName)
-	default:
-		return fmt.Errorf("language %q not supported. %s", d.lang, LanguageNotSupportedError)
-	}
-	if err != nil {
-		d.logger.Warnf(ctx, "Unable to check for resource availability (%s). Moving on...", err.Error())
-		return nil
-	}
-	if len(resourceNames) == 0 {
-		// Turbine bug or misconfigured app.
-		d.logger.Warnf(ctx, "Unable to check for resource availability. Moving on...")
-		return nil
-	}
-
+func (d *Deploy) getResourceCheckErrorMessage(ctx context.Context, resourceNames []string) string {
 	var errStr string
 	for _, name := range resourceNames {
 		resource, err := d.client.GetResourceByNameOrID(ctx, name)
 		if err != nil {
 			if errStr != "" {
 				errStr += "; "
-			} else {
-				errStr = "run 'meroxa resource list' to double check the name and try again; "
 			}
+
 			if strings.Contains(err.Error(), "could not find") {
 				errStr += fmt.Sprintf("could not find resource %q", name)
 			} else {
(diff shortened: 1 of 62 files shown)


cmd/meroxa/root/apps/deploy.go
```diff
@@ -524,6 +526,69 @@ func (d *Deploy) validate(ctx context.Context) error {
...
+	switch {
+	case resourceNames == nil, len(resourceNames) == 0:
+		return errors.New("no resources defined in your Turbine app")
+	case len(resourceNames) < requiredNumberOfResources:
+		return fmt.Errorf(
+			"insufficient number of resources defined: %v."+
+				"Please add at least one source and one destination resource to your app definition",
+			resourceNames,
+		)
+	}
```
ahmeroxa, line 559: I don't think we should include this check. I have a Turbine app that uses a Postgres source but uses only a function to push data into Salesforce (i.e. no destination).

cmd/meroxa/root/apps/deploy.go
```diff
@@ -42,9 +42,10 @@ import (
 )
 
 const (
-	dockerHubUserNameEnv    = "DOCKER_HUB_USERNAME"
-	dockerHubAccessTokenEnv = "DOCKER_HUB_ACCESS_TOKEN" // nolint:gosec
-	pollDuration            = 2 * time.Second
+	dockerHubUserNameEnv      = "DOCKER_HUB_USERNAME"
+	dockerHubAccessTokenEnv   = "DOCKER_HUB_ACCESS_TOKEN" // nolint:gosec
+	pollDuration              = 2 * time.Second
+	requiredNumberOfResources = 2
```
ahmeroxa, line 48: Mentioned below, but I think we should remove this check.

jayjayjpg: Sounds good!

ahmeroxa (approved):
LGTM 👍

cmd/meroxa/root/apps/deploy.go
```diff
@@ -524,6 +525,62 @@ func (d *Deploy) validate(ctx context.Context) error {
...
+		resourceNames, err = turbinePY.GetResourceNames(ctx, d.logger, d.path, d.appName)
+	default:
+		return fmt.Errorf("language %q not supported. %s", d.lang, LanguageNotSupportedError)
+	}
+
+	if err != nil {
+		return fmt.Errorf("unable to read resource definition from app: %s", err.Error())
+	}
+
+	if resourceNames == nil || len(resourceNames) == 0 {
```
ahmeroxa, line 549:
```suggestion
	if len(resourceNames) == 0 {
```

jayjayjpg: Thank you!

cmd/meroxa/root/apps/deploy.go
```diff
@@ -524,6 +525,62 @@ func (d *Deploy) validate(ctx context.Context) error {
...
+			}
+		} else if resource.Status.State != meroxa.ResourceStateReady {
+			if errStr != "" {
+				errStr += "; "
+			}
+			errStr += fmt.Sprintf("resource %q is not ready and usable", resource.Name)
+		}
+	}
+	if errStr != "" {
+		errStr += ";\n\n ⚠️  Run 'meroxa resource list' to verify that the names of resources" +
```
raulb, line 575: This needs a space, otherwise it looks like the following: .


```suggestion
		errStr += ";\n\n ⚠️  Run 'meroxa resource list' to verify that the names of resources " +
```

cmd/meroxa/root/apps/deploy.go
```diff
@@ -524,6 +525,62 @@ func (d *Deploy) validate(ctx context.Context) error {
 	return turbineCLI.ValidateBranch(ctx, d.logger, d.path)
 }
 
+func (d *Deploy) checkResourceAvailability(ctx context.Context) error {
+	d.logger.Infof(ctx, "Checking resource availability for application %q (%s) before deployment...", d.appName, d.lang)
```
raulb, line 529: we could probably leverage the use of `d.logger.StartSpinner` here to indicate it that way, what do you think? @jayjayjpg

cmd/meroxa/root/apps/deploy.go
```diff
@@ -524,6 +525,62 @@ func (d *Deploy) validate(ctx context.Context) error {
...
+
+	var resourceNames []string
+	var err error
+
+	switch d.lang {
+	case GoLang:
+		resourceNames, err = turbineGo.GetResourceNames(ctx, d.logger, d.path, d.appName)
+	case JavaScript, NodeJs:
```
raulb, line 537: I don't think we need to case for `NodeJs` since we standarize languages in the validation step: https://github.com/meroxa/cli/blob/master/cmd/meroxa/root/apps/deploy.go#L483.

cmd/meroxa/root/apps/deploy.go
```diff
@@ -524,6 +525,62 @@ func (d *Deploy) validate(ctx context.Context) error {
...
+	}
+
+	if len(resourceNames) == 0 {
+		return errors.New("no resources defined in your Turbine app")
+	}
+
+	var errStr string
+
+	for _, name := range resourceNames {
+		resource, err := d.client.GetResourceByNameOrID(ctx, name)
```
raulb, line 556: I would probably extract all this logic to a separate function and test this to make sure we generate the error string we'd expect based on the different errors that could be returned by the API.

cmd/meroxa/root/apps/deploy.go
```diff
@@ -524,6 +525,62 @@ func (d *Deploy) validate(ctx context.Context) error {
...
+	var err error
+
+	switch d.lang {
+	case GoLang:
+		resourceNames, err = turbineGo.GetResourceNames(ctx, d.logger, d.path, d.appName)
+	case JavaScript, NodeJs:
+		resourceNames, err = turbineJS.GetResourceNames(ctx, d.logger, d.path, d.appName)
+	case Python:
+		resourceNames, err = turbinePY.GetResourceNames(ctx, d.logger, d.path, d.appName)
+	default:
```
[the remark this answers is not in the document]
raulb: I just noticed we stop the spinner here https://github.com/meroxa/cli/pull/352/files#diff-9660bb3708bfbd2d8adc60d0b68dfd003cbf13421ea4212747439a8c11fc54b5R580, but where are we starting it?

raulb (changes requested):
@jayjayjpg looking good so far! I think this pull-request would benefit from tests. I also left some other comments. Let me know what you think!

cmd/meroxa/root/apps/deploy.go
jayjayjpg: Thank you! :sparkles:

cmd/meroxa/root/apps/deploy.go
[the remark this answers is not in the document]
jayjayjpg: Good catch, I think with the suggestion you made [here](https://github.com/meroxa/cli/pull/352#discussion_r885872140) this should work!

cmd/meroxa/root/apps/deploy.go
```diff
@@ -524,6 +525,68 @@ func (d *Deploy) validate(ctx context.Context) error {
...
+	}
+
+	if len(resourceNames) == 0 {
+		return errors.New("no resources defined in your Turbine app")
+	}
+
+	errStr := d.getResourceCheckErrorMessage(ctx, resourceNames)
+
+	if errStr != "" {
+		errStr += ";\n\n ⚠️  Run 'meroxa resource list' to verify that the names of resources " +
```
raulb, line 580: Nitpicking. When listing (dealing with with a plural list), I tend to use the plural form `meroxa resources list`. 

```suggestion
		errStr += ";\n\n ⚠️  Run 'meroxa resources list' to verify that the resource names " +
```

cmd/meroxa/root/apps/deploy.go
```diff
@@ -524,6 +525,68 @@ func (d *Deploy) validate(ctx context.Context) error {
...
+			"defined in your Turbine app are identical to the resources you have created on the Meroxa platform before deploying again"
```
raulb, line 581: Nitpicking (again). When mentioning Meroxa products, I tend to use the uppercase form (i.e.: `Platform` instead of `platform`.

```suggestion
			"defined in your Turbine app are identical to the resources you have created on the Meroxa Platform before deploying again"
```

cmd/meroxa/root/apps/deploy.go
```diff
@@ -524,6 +525,68 @@ func (d *Deploy) validate(ctx context.Context) error {
...
+		d.logger.StopSpinnerWithStatus("Resource availability check failed", log.Failed)
+		return fmt.Errorf("%s", errStr)
+	}
+
+	d.logger.StopSpinnerWithStatus("Found all required resources!", log.Successful)
```
raulb, line 586: What do you think about this change? Trying to make it more explicit. 

```suggestion
	d.logger.StopSpinnerWithStatus("Can access to your Turbine resources", log.Successful)
```

cmd/meroxa/root/apps/deploy.go
```diff
@@ -554,7 +623,7 @@ func (d *Deploy) rmBinary() {
 }
 
 // tearDownExistingResources will only delete the application and its associated entities if it hasn't been created
-// or whether it's in a non runnin state.
+// or whether it's in a non-running state.
```
raulb, line 626: Thank you for catching this typo!

cmd/meroxa/turbine_cli/utils.go
```diff
@@ -20,7 +20,7 @@ import (
 	"github.com/meroxa/cli/log"
 )
 
-const turbineJSVersion = "0.1.7"
+const turbineJSVersion = "0.1.8"
```
raulb, line 23: Definitely not part of this PR, but we should find a way to automate this.

utils/tests.go
```diff
@@ -33,18 +33,28 @@ func GeneratePipelineWithEnvironment() meroxa.Pipeline {
 	return p
 }
 
-func GenerateResource() meroxa.Resource {
-	return meroxa.Resource{
+func GenerateResource(resourceName, resourceState string) meroxa.Resource {
```
raulb, line 36: I like the versatility of including this option. 

I feel like I haven't done this in the past in this repository, so I definitely understand you taking the approach of sending empty strings in those cases we didn't care about the name of state, but maybe it'd be more idiomatic to define another function that doesn't pass name or state, that could call this one with those empty strings values.

i.e.:

Define:

```go
func GenerateResource() meroxa.Resource {
  return GenerateResourceWithNameAndStatus("", "")
}
```

And still call `GenerateResource()` when we don't mind using a specific name or state.

raulb (approved):
@jayjayjpg great work! I only left some non-blocking comments. Let me know what you think, but this is definitely ready to ship as is.

cmd/meroxa/root/apps/deploy.go
jayjayjpg: Thank you for the suggestion! :sparkles:

cmd/meroxa/turbine_cli/utils.go
jayjayjpg: Agreed! I think I can borrow some of your work on updating the turbine-go dependency in the CLi here as well

utils/tests.go
jayjayjpg: That's much nicer to read indeed and I can refactor this accordingly!

cmd/meroxa/root/apps/deploy.go
jayjayjpg: Thank you for the suggestion!

raulb:
@jayjayjpg let me know if you need help releasing this to a newer version of our CLI and if this needs to be coordinated with a changelog (happy also to help on that).

PR #368, opened 2022-05-31
[WIP] upd(deploy): reword predeployment user messaging to be more descriptive

jayjayjpg:
## Description of change

This is just a quick pass on how the wording for the binary build step during app deployments could look like. Feedback and suggestions are welcome!

This is a follow-up [to this demo and discussion](https://meroxa.slack.com/archives/C02L14X6TA9/p1653672323863129)

## Type of change

- [ ]  New feature
- [x]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

## How was this tested?

- [ ]  Unit Tests
- [x]  Tested in staging
- [ ]  Tested in minikube

## Demo

| before | after |
|--------|-------|
|appdeploys-before|appdeploys-after|

### DIFF
--- a/cmd/meroxa/root/apps/deploy.go
+++ b/cmd/meroxa/root/apps/deploy.go
@@ -408,7 +408,7 @@ func (d *Deploy) buildApp(ctx context.Context) error {
 	var err error
 
 	// Without the " " at the beginning of `suffix`, spinner looks next to word (only on this occurrence)
-	d.logger.StartSpinner("\t", " Building application...")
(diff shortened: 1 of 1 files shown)


raulb (changes requested):
@jayjayjpg This would look right for go apps, but I don't think would make sense for JS/Python applications since there's no binary involved. Hence why I went for a generic "Building application" in the first place.

jayjayjpg:
@raulb Got it, I haven't considered that before and thank you for sharing!

PR #371, opened 2022-06-01
Allow for skipping of git validations

dianadoherty:
## Description of change

Allow for skipping of git validations

Fixes https://github.com/meroxa/acceptance/issues/101

## Type of change

- [x]  New feature
- [ ]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

## How was this tested?

- [ ]  Unit Tests
- [x]  Tested in staging
- [ ]  Tested in minikube

## Demo

| before | after |
|--------|-------|
|||

### DIFF
--- a/cmd/meroxa/root/apps/deploy.go
+++ b/cmd/meroxa/root/apps/deploy.go
@@ -42,6 +42,8 @@ import (
 )
 
 const (
+	deployGitSha            = "DEPLOY_GIT_SHA"
+	deploySkipGitValidation = "DEPLOY_SKIP_GIT_VALIDATION"
 	dockerHubUserNameEnv    = "DOCKER_HUB_USERNAME"
 	dockerHubAccessTokenEnv = "DOCKER_HUB_ACCESS_TOKEN" // nolint:gosec
 	pollDuration            = 2 * time.Second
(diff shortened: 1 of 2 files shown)


cmd/meroxa/root/apps/deploy.go
```diff
@@ -113,6 +115,20 @@ func (d *Deploy) Flags() []builder.Flag {
...
+func (d *Deploy) getDeployGitSha() string {
+	if v := os.Getenv(deployGitSha); v != "" {
+		return v
+	}
+	return d.config.GetString(deployGitSha)
+}
+
+func (d *Deploy) skipGitValidation() bool {
+	if v := os.Getenv(deploySkipGitValidation); v != "" {
+		return v == "true"
```
jayjayjpg, line 127: Just asking out of curiosity -- could we also directly return `true` here or would this be different from `v == "true"`?

jayjayjpg: Oh I think I got it now, the `v` will always return a stringified version of the boolean `true` and this is just making sure that the environment variable is set correctly -- nevermind me then! :-)

jayjayjpg (approved):
This reads great! Thank you for the swift iteration on this feature! :sparkles:

neovintage:
@dianadoherty can any customer allow this to happen?

jayjayjpg:
@neovintage @raulb @dianadoherty If I understood the code correctly, this is only a feature that's only accessible to users who intentionally choose to disable git validation via a feature flag, no? The user flow would remain unchanged for any other end user by default

dianadoherty:
> @neovintage @raulb @dianadoherty If I understood the code correctly, this is only a feature that's only accessible to users who intentionally choose to disable git validation via a feature flag, no? The user flow would remain unchanged for any other end user by default

We had a discussion about this PR in the funtime sync. I'll close out this ticket since a different approach was taken to solve this problem :)

PR #405, opened 2022-08-09
Turbine-py feature branch changes (--listresources and app name)

anna-cross:
## Description of change

Changes as part of feature branch project for turbine-py. List resources on turbine-py now dumbs json similar to go, the code for parsing is now the same as for golang too. 

Deploy also now requires an app name, this will be using to override app.json app_name. 

Fixes (https://github.com/meroxa/product/issues/398)

Relevant PRs: 

https://github.com/meroxa/turbine-py/pull/80
https://github.com/meroxa/meroxa-py/pull/34

## Type of change

- [x]  Refactor

## How was this tested?

- [ ]  Unit Tests
- [x]  Tested in staging

## Demo

| before | after |
|--------|-------|

### DIFF
--- a/cmd/meroxa/turbine_cli/python/deploy.go
+++ b/cmd/meroxa/turbine_cli/python/deploy.go
@@ -57,7 +57,7 @@ func NeedsToBuild(path string) (bool, error) {
 // RunDeployApp creates Application entities.
 func RunDeployApp(ctx context.Context, l log.Logger, path, imageName, appName, gitSha string) error {
 	cmd := exec.Command("turbine-py", "clideploy", path, imageName, appName, gitSha)
-  
+
 	accessToken, _, err := global.GetUserToken()
 	if err != nil {
 		return err
--- a/cmd/meroxa/turbine_cli/python/deploy.go
+++ b/cmd/meroxa/turbine_cli/python/deploy.go
@@ -89,7 +89,7 @@ func GetResourceNames(ctx context.Context, l log.Logger, appPath, appName string
 		return names, errors.New(string(output))
 	}
 
-	var parsed []map[string]interface{}
+	var parsed []turbinecli.ApplicationResource
 	if err := json.Unmarshal(output, &parsed); err != nil {
 		// fall back if not json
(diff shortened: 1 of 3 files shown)


cmd/meroxa/turbine_cli/python/deploy.go
```diff
@@ -87,16 +88,32 @@ func GetResourceNames(ctx context.Context, l log.Logger, appPath, appName string
 	if err != nil {
 		return names, errors.New(string(output))
 	}
-	r := regexp.MustCompile("^turbine-response: \\[(.*)\\]\n")
-	match := r.FindStringSubmatch(string(output))
-	if match == nil || len(match) < 2 {
-		return names, fmt.Errorf("unable to verify resource availability for Meroxa Application at %s; %s", appPath, string(output))
+
+	var parsed []map[string]interface{}
```
janelletavares, line 92: what does this data actually look like? i'm a little concerned about this data type

anna-cross: The listresources response looks like this: 

[ { "name": "source", "source": true, "destination": false, "collection": "orders" }, { "name": "dest", "source": false, "destination": true, "collection": "test" } ]

It is now the same format as turbine-go response , the code for parsing it is essentially the same now as turbine go too

lyuboxa: @anna-cross we can create a struct type to formalize it, this will help us decode what we expect

lyuboxa: @janelletavares likely came from here https://github.com/meroxa/cli/blob/master/cmd/meroxa/turbine_cli/golang/deploy.go#L52-L56

i think once we have this in place we can refactor it to be stricter.

janelletavares: +1 would prefer a []struct for the unmarshling pointer

anna-cross: Updated to use a struct :D will do the same for JS

lyuboxa (approved):

PR #408, opened 2022-08-10
Meroxa go upgrade headers

nassor:
## Description of change

Updating `meroxa-go` package to be able to inject headers in the `apiClient.MakeRequest` calls.

_Extra: GOPATH was required for the app's initialization, and `build.Default.GOPATH` was added as a fallback._

## Type of change

- [X] New feature
- [X] Bugfix

## How was this tested?

- [X]  Unit Tests

### DIFF
--- a/cmd/meroxa/turbine_cli/utils.go
+++ b/cmd/meroxa/turbine_cli/utils.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"go/build"
 	"io"
 	"os"
 	"os/exec"
@@ -237,11 +238,15 @@ func GetGitSha(appPath string) (string, error) {
 	return string(output), nil
 }
 
-func GoInit(l log.Logger, appPath string, skipInit, vendor bool) error {
+func GoInit(ctx context.Context, l log.Logger, appPath string, skipInit, vendor bool) error {
 	l.StartSpinner("\t", "Running golang module initializing...")
 	skipLog := "skipping go module initialization\n\tFor guidance, visit " +
 		"https://docs.meroxa.com/beta-overview#go-mod-init-for-a-new-golang-turbine-data-application"
 	goPath := os.Getenv("GOPATH")
+	if goPath == "" {
+		goPath = build.Default.GOPATH
+		l.Infof(ctx, "using default GOPATH: %s", goPath)
+	}
 	if goPath == "" {
 		l.StopSpinnerWithStatus("$GOPATH not set up; "+skipLog, log.Warning)
 		return nil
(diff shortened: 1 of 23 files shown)


cmd/meroxa/turbine_cli/utils.go
```diff
@@ -242,6 +243,9 @@ func GoInit(l log.Logger, appPath string, skipInit, vendor bool) error {
 	skipLog := "skipping go module initialization\n\tFor guidance, visit " +
 		"https://docs.meroxa.com/beta-overview#go-mod-init-for-a-new-golang-turbine-data-application"
 	goPath := os.Getenv("GOPATH")
+	if goPath == "" {
+		goPath = build.Default.GOPATH
```
janelletavares, line 247: probably good to log here that the default is being used here because GOPATH is not set

samirketema: agreed - would be nice to display a log when it happens

nassor: Solved.

samirketema (approved):
lgtm! 🎉

cmd/meroxa/turbine_cli/utils.go
janelletavares: did you test this manually?  i don't think that's compatible with the spinner

janelletavares: Screenshot from 2022-08-11 12-08-21

cmd/meroxa/turbine_cli/utils.go
```diff
@@ -237,11 +238,15 @@ func GetGitSha(appPath string) (string, error) {
...
+func GoInit(ctx context.Context, l log.Logger, appPath string, skipInit, vendor bool) error {
 	l.StartSpinner("\t", "Running golang module initializing...")
 	skipLog := "skipping go module initialization\n\tFor guidance, visit " +
 		"https://docs.meroxa.com/beta-overview#go-mod-init-for-a-new-golang-turbine-data-application"
 	goPath := os.Getenv("GOPATH")
+	if goPath == "" {
+		goPath = build.Default.GOPATH
+		l.Infof(ctx, "using default GOPATH: %s", goPath)
+	}
 	if goPath == "" {
```
raulb, line 250: @nassor looks like this conditional will never take place (if `goPath` was empty, then it was set to `build.Default.GOPATH`.  We should remove this one also stopping the spinner since we started it on[ line 242](https://github.com/meroxa/cli/pull/408/files#diff-26b7e43c725b91b24ee83f56a2e8a9f0e595b868fb1d50d33c9bff65592a15b9R242).

raulb (commented):
@nassor there is a minor change we should make here.

cmd/meroxa/turbine_cli/utils.go
raulb: @nassor ignore me. I just saw your follow up PR https://github.com/meroxa/cli/pull/409 👍

PR #413, opened 2022-08-15
Refactor turbine cli utils

lyuboxa:
## Description of change

Use go commands native way to switch to a folder when executing a command. 
Add functionality stub to prepare app name. Isolate validation to determine if the a branch is either `main` or `master`.

Part of https://github.com/meroxa/product/issues/502

## Type of change

- [x]  New feature
- [ ]  Bug fix
- [x]  Refactor
- [ ]  Documentation

## How was this tested?

- [ ]  Unit Tests
- [ ]  Tested in staging
- [ ]  Tested in minikube

## Demo

| before | after |
|--------|-------|
|||

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,11 @@
-SHELL=/bin/bash -o pipefail
-GIT_COMMIT:=$(shell git rev-parse --short HEAD)
-LDFLAGS:=-X main.GitCommit=${GIT_COMMIT}
-GIT_UNTRACKED=$(shell git diff-index --quiet HEAD -- || echo "(updated)")
-LDFLAGS+=-X main.GitUntracked=${GIT_UNTRACKED}
-GIT_TAG=$(shell git describe)
-LDFLAGS+=-X main.GitLatestTag=${GIT_TAG}
+SHELL           = /bin/bash -o pipefail
+GIT_COMMIT     := $(shell git rev-parse --short HEAD)
+LDFLAGS        := -X main.GitCommit=${GIT_COMMIT}
+GIT_UNTRACKED   = $(shell git diff-index --quiet HEAD -- || echo "(updated)")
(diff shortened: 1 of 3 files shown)


dianadoherty (approved):

lyuboxa:
Acceptance tests on staging are flaking and this change is not affecting this.

Makefile
```diff
@@ -1,10 +1,11 @@
...
-GIT_TAG=$(shell git describe)
-LDFLAGS+=-X main.GitLatestTag=${GIT_TAG}
+SHELL           = /bin/bash -o pipefail
+GIT_COMMIT     := $(shell git rev-parse --short HEAD)
+LDFLAGS        := -X main.GitCommit=${GIT_COMMIT}
+GIT_UNTRACKED   = $(shell git diff-index --quiet HEAD -- || echo "(updated)")
+LDFLAGS        += -X main.GitUntracked=${GIT_UNTRACKED}
+GIT_TAG         = $(shell git describe)
+LDFLAGS        += -X main.GitLatestTag=${GIT_TAG}
+REBUILD_DOCS    ?= true
```
raulb, line 8: @lyuboxa why was this necessary?

raulb: AFACIT, this should be `"true"` to make sure we still can automatically generate documentation.

raulb: Opened PR to fix and add documentation https://github.com/meroxa/cli/pull/415

ahmeroxa: `?= true` should default to true unless it's set, so I don't think there should be a net change in behavior unless someone explicitly sets it to false.

raulb: @ahmeroxa AFAICT, it's not the same as `"true"` https://github.com/meroxa/cli/pull/413/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R28. So default behavior is "don't" build docs.

PR #421, opened 2022-08-23
Deploy app when skip flag provided for non unique destination collections

dianadoherty:
## Description of change

Allow for app deploys when skip flag provided for non unique destination collections

Fixes https://github.com/meroxa/product/issues/510

## Type of change

- [x]  New feature
- [ ]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

## How was this tested?

- [x]  Unit Tests
- [ ]  Tested in staging
- [ ]  Tested in minikube

## Demo

Screen Shot 2022-08-23 at 2 46 43 PM

### DIFF
--- a/cmd/meroxa/root/apps/deploy.go
+++ b/cmd/meroxa/root/apps/deploy.go
@@ -62,10 +62,11 @@ type deployApplicationClient interface {
 
 type Deploy struct {
 	flags struct {
-		Path                 string `long:"path" usage:"path to the app directory (default is local directory)"`
+		Path                 string `long:"path" usage:"Path to the app directory (default is local directory)"`
 		DockerHubUserName    string `long:"docker-hub-username" usage:"DockerHub username to use to build and deploy the app" hidden:"true"`         //nolint:lll
 		DockerHubAccessToken string `long:"docker-hub-access-token" usage:"DockerHub access token to use to build and deploy the app" hidden:"true"` //nolint:lll
 		Spec                 string `long:"spec" usage:"Deployment specification version to use to build and deploy the app" hidden:"true"`
+		SkipUniqueCollection bool   `long:"skip-unique-collection" usage:"Skips unique destination collection validation" hidden:"true"`
 	}
 
 	client        deployApplicationClient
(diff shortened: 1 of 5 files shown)


cmd/meroxa/root/apps/deploy.go
```diff
@@ -62,10 +62,11 @@ type deployApplicationClient interface {
 
 type Deploy struct {
 	flags struct {
-		Path                 string `long:"path" usage:"path to the app directory (default is local directory)"`
+		Path                 string `long:"path" usage:"Path to the app directory (default is local directory)"`
```
raulb, line 65: 👍 thank you!

cmd/meroxa/root/apps/deploy.go
```diff
@@ -62,10 +62,11 @@ type deployApplicationClient interface {
...
 		DockerHubUserName    string `long:"docker-hub-username" usage:"DockerHub username to use to build and deploy the app" hidden:"true"`         //nolint:lll
 		DockerHubAccessToken string `long:"docker-hub-access-token" usage:"DockerHub access token to use to build and deploy the app" hidden:"true"` //nolint:lll
 		Spec                 string `long:"spec" usage:"Deployment specification version to use to build and deploy the app" hidden:"true"`
+		SkipUniqueCollection bool   `long:"skip-unique-collection" usage:"Skips unique destination collection validation" hidden:"true"`
```
raulb, line 69: If the intention of this flag will be only for internal purposes, I agree with being `hidden`, and its naming. However, I think we should reconsider it if we think this is something external users will take advantage since I don't think it's very intuitive.

raulb (approved):
👍 LGTM

cmd/meroxa/root/apps/deploy.go
dianadoherty: Yes, this flag will be used for external users. Right now this feature is blocked under a feature flag, so I want to unhide it as soon as we're ready to have it go live :) 

What do you recommend renaming the flag to?

justmisosoup: @raulb This definitely isn't the ideal nor final solution. The problem we were aiming to solve was preventing customers from accidentally deploying data apps from feature branches with downstream resources in common to their production apps. The --skip flag was intended to give customers a way to work around some validations that they may intend to bypass. 

I think there are more intuitive solutions that can be explored with future work we have in the backlog and with IR in place.

cc: @dianadoherty

dianadoherty: I'll go ahead and merge this PR and create a draft PR to unhide this flag. We can also rename it there if need be :)

raulb: Sounds good. Let's keep iterating on this before we make this flag visible and need to document it!

PR #433, opened 2022-09-06
[IR] accommodate preIR and IR deployments side by side

janelletavares:
## Description of change

Fixes https://github.com/meroxa/product/issues/544

## Type of change

- [x]  New feature
- [ ]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

## How was this tested?

- [x]  Unit Tests
- [ ]  Tested in staging
- [ ]  Tested in minikube

## Demo

| before | after |
|--------|-------|
|||

### DIFF
--- a/cmd/meroxa/root/apps/deploy.go
+++ b/cmd/meroxa/root/apps/deploy.go
@@ -610,10 +610,8 @@ func (d *Deploy) prepareAppName(ctx context.Context) string {
 
 func (d *Deploy) waitForDeployment(ctx context.Context) error {
 	startTime := time.Now().UTC()
-	finished := false
 	logs := []string{}
-	deploymentErr := fmt.Errorf("timed out; check `apps logs`")
-	for !finished && (time.Now().UTC().Sub(startTime) < 5*time.Minute) {
+	for time.Now().UTC().Sub(startTime) < 5*time.Minute {
 		var deployment *meroxa.Deployment
 		deployment, err := d.client.GetLatestDeployment(ctx, d.appName)
 		if err != nil {
@@ -622,23 +620,27 @@ func (d *Deploy) waitForDeployment(ctx context.Context) error {
 		newLogs := strings.Split(deployment.OutputLog.String, "\n")
 		if len(newLogs) > len(logs) {
 			for i := len(logs); i < len(newLogs); i++ {
-				if len(logs) != 0 && i != 0 {
+				if len(logs) == 0 && i != 0 {
+					d.logger.StopSpinner(newLogs[i-1])
+				} else if len(logs) != 0 {
 					d.logger.StopSpinner(newLogs[i-1])
 				}
 				d.logger.StartSpinner("\t", newLogs[i])
 			}
-			logs = append(logs, newLogs...)
+			logs = newLogs
 		}
 		if deployment.Status.State == meroxa.DeploymentStateDeployed {
-			finished = true
-			deploymentErr = nil
+			l := len(logs)
+			d.logger.StopSpinner(logs[l-1])
+			return nil
 		} else if deployment.Status.State == meroxa.DeploymentStateDeployingError ||
 			deployment.Status.State == meroxa.DeploymentStateRollingBackError {
-			d.logger.StopSpinnerWithStatus(logs[len(logs)-1], log.Failed)
+			l := len(logs)
+			d.logger.StopSpinnerWithStatus(logs[l-1], log.Failed)
 			return fmt.Errorf("failed to deploy Application %q", d.appName)
 		}
 	}
-	return deploymentErr
+	return fmt.Errorf("timed out; check `apps logs`")
 }
 
 func hasFeatureFlag(f string) bool {
--- a/cmd/meroxa/root/apps/deploy.go
+++ b/cmd/meroxa/root/apps/deploy.go
@@ -689,6 +689,7 @@ func (d *Deploy) Execute(ctx context.Context) error {
 		return err
 	}
 
+	d.specVersion = d.flags.Spec
 	if d.specVersion == "" {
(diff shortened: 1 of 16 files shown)


cmd/meroxa/root/apps/deploy.go
```diff
@@ -649,25 +638,54 @@ func (d *Deploy) Execute(ctx context.Context) error {
...
 		return err
 	}
 
-	return d.deployApp(ctx, d.fnName, gitSha, d.specVersion)
+	if d.specVersion == "" {
+		app, err = d.checkPreIRApplication(ctx)
+		if err != nil {
+			return err
+		}
+	}
```
janelletavares, line 684: @TODO get the latest deployment status, potentially in a loop if it's not implemented as a stream

cmd/meroxa/root/apps/deploy.go
```diff
@@ -649,25 +638,54 @@ func (d *Deploy) Execute(ctx context.Context) error {
...
-	if err := d.tearDownExistingResources(ctx); err != nil {
-		return err
+	var app *meroxa.Application
+	if d.specVersion == "" {
+		// ⚠️ This is only until we re-deploy applications applying only the changes made
+		if err := d.tearDownExistingResources(ctx); err != nil {
+			return err
+		}
+	} else {
+		// Intermediate Representation Workflow
```
janelletavares, line 653: @TODO check feature flag

janelletavares:
I tested the pre IR and IR code paths for go apps.  Test results pending for py and js.

cmd/meroxa/root/apps/deploy.go
```diff
@@ -238,22 +242,24 @@ func (d *Deploy) getPlatformImage(ctx context.Context) (string, error) {
...
-	if err != nil {
-		return err
+	if specVersion != "" {
+		input := &meroxa.CreateDeploymentInput{
+			Application: meroxa.EntityIdentifier{Name: null.StringFrom(d.appName)},
+			GitSha:      gitSha,
+			SpecVersion: null.StringFrom(specVersion),
+			Spec:        null.StringFrom(deploymentSpec),
+		}
+		_, err = d.client.CreateDeployment(ctx, input)
```
raulb, line 256: I thought this was going to be done from the API. 

Are we planning on making the API create the deployment in some point? I believe it should be more stable to leave the API the responsibility of doing that.

cmd/meroxa/root/apps/deploy.go
```diff
@@ -600,7 +606,45 @@ func (d *Deploy) prepareAppName(ctx context.Context) string {
 	return appName
 }
 
+func (d *Deploy) waitForDeployment(ctx context.Context) error {
+	startTime := time.Now().UTC()
+	logs := []string{}
+	for time.Now().UTC().Sub(startTime) < 5*time.Minute {
```
raulb, line 612: Is this running every ms or how often?

cmd/meroxa/root/apps/deploy.go
lyuboxa: I thought creating the deployment which carries the spec from the CLI was the way? Since this is what carries the IR?

cmd/meroxa/root/apps/deploy.go
```diff
@@ -600,7 +606,45 @@ func (d *Deploy) prepareAppName(ctx context.Context) string {
...
+		var deployment *meroxa.Deployment
+		deployment, err := d.client.GetLatestDeployment(ctx, d.appName)
+		if err != nil {
+			return err
+		}
+		newLogs := strings.Split(deployment.OutputLog.String, "\n")
```
lyuboxa, line 618: What is output log? I see it in meroxa-go log but the platform doesn't do anything with that. I'd rather we don't use this until it is implemented

cmd/meroxa/root/apps/deploy.go
lyuboxa: can we use a ticket with a context timeout instead?
kind of looks like this
```
	cctx, cancel := context.WithTimeout(ctx, <timeout>)
	defer cancel()

	t := time.NewTicker(1 * time.Second)
	defer t.Stop()

	for {
		select {
		case <-t.C:
			<do work>
		case <-cctx.Done():
                       <timeout expired>
			return cctx.Err()
		}
	}
```

cmd/meroxa/root/apps/deploy.go
janelletavares: I don't understand the comment, Raul.  Could you rephrase?  What exactly did you think was going to be done in the API?  Did you think the application endpoint would create a populated deployment?  The Application endpoint will create an empty pipeline in v2, and the v1 application endpoint backfills an empty deployment for already-populated applications.

cmd/meroxa/root/apps/deploy.go
janelletavares: yeah, it doesn't sleep or backoff at all, Raul, a small sleep might be good.  Given that deployments are immediately done with no output logs, this doesn't do much of anything now.  mostly exercised by the unit test.

cmd/meroxa/root/apps/deploy.go
janelletavares: the worker doesn't add any outlogs right now, but in the future i would expect one log per step. we want the logs to roughly match to the CLI logs today

jayjayjpg (approved):
This reads great, thank you for putting all of this together so quickly! :sparkles:

cmd/meroxa/root/apps/deploy.go
janelletavares: this is the result of the second RFC.  this is what we do if we make sure the application exists before the deployment.  we create deployments and not applications after the application exists.

PR #486, opened 2022-10-17
renaming usage and docs

### LINKED ISSUE #485: [Collaboration] Accounts to Projects rename in CLI
## Description

Following a decision to change language from **Accounts** to **Projects** for **Collaborations**, we will be changing external communication to customers including the Command Line Interface (CLI).

## Out of scope

We will not be making back-end changes.

nassor:
## Description of change

Rename Accounts to Projects to fit into Meroxa information architecture

Fixes #485 

## Type of change

- [X]  Refactor

### DIFF
--- a/cmd/meroxa/root/account/accounts.go
+++ b/cmd/meroxa/root/account/accounts.go
@@ -23,7 +23,7 @@ import (
 )
 
 // Account implement the projects management onto the Meroxa platform.
-// Accounts were renames Projects, renamed to fix the information architecture.
+// Accounts on the platform were renamed to Projects, to fix the information architecture.
 type Account struct{}
 
 var (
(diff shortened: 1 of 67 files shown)


cmd/meroxa/root/account/accounts.go
```diff
@@ -22,6 +22,8 @@ import (
 	"github.com/meroxa/cli/cmd/meroxa/builder"
 )
 
+// Account implement the projects management onto the Meroxa platform.
+// Accounts were renames Projects, renamed to fix the information architecture.
```
samirketema, line 26: @nassor Thanks for the context. I think this reads a bit clearer / more concise with the following change. What do you think? 
```suggestion
// Accounts on the platform were renamed to Projects, to fix the information architecture.
```

samirketema (approved):
Looks good to me - just one suggestion on clarifying a comment

PR #504, opened 2022-10-25
bug: Nil pointer in upload

janelletavares:
## Description of change

Fixes <GitHub Issue>

## Type of change

- [ ]  New feature
- [x]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

## How was this tested?

- [ ]  Unit Tests
- [ ]  Tested in staging
- [x]  Tested in minikube

## Demo

| before | after |
|--------|-------|
|||

### DIFF
--- a/cmd/meroxa/turbine/utils.go
+++ b/cmd/meroxa/turbine/utils.go
@@ -472,7 +472,7 @@ func uploadFile(ctx context.Context, logger log.Logger, filePath, url string) er
 		return err
 	}
 
-	if res.Body != nil {
+	if res != nil && res.Body != nil {
 		defer res.Body.Close()
 	}
 


cmd/meroxa/turbine/utils.go
```diff
@@ -472,7 +472,7 @@ func uploadFile(ctx context.Context, logger log.Logger, filePath, url string) er
 		return err
 	}
 
-	if res.Body != nil {
+	if res != nil && res.Body != nil {
```
samirketema, line 475: This looks fine, but seems like we need to add test coverage for this function and/or `UploadSource()`. Thoughts on filing a KTLO issue for this?

samirketema (approved):

nassor (approved):
I thought that was weird first, but it can actually happen: https://cs.opensource.google/go/go/+/refs/tags/go1.19.2:src/net/http/client.go;l=684-689;drc=55eaae452cf69df768b2aaf6045db22d6c1a4029;bpv=1;bpt=1

cmd/meroxa/turbine/utils.go
janelletavares: good call out. adding unit tests for this file was the remaining task for this ticket https://github.com/meroxa/turbine-project/issues/251

most of the functions in this file will be hard to test because of external calls. with dependency injection or something like that, we'll see what we can do.

PR #508, opened 2022-10-26
feat: make ir deployments default

raulb:
## Description of change

Fixes https://github.com/meroxa/product/issues/559

## Type of change

- [x]  New feature
- [ ]  Bug fix
- [x]  Refactor
- [ ]  Documentation

## How was this tested?

- [x]  Unit Tests
- [ ]  Tested in staging
- [ ]  Tested in minikube

## Demo

**Before**

Regular deployment (pre: intermediate representation)

```
$ meroxa apps deploy .
```

To use a deployment using Intermediate representation:

```
$ meroxa apps deploy . --spec 0.1.1
```

**After**

To use a deployment using Intermediate representation, it'll fetch latest version and it'll be sent to all turbine libraries.

```
$ meroxa apps deploy .
```

## Additional references

This will be a nice to have right after we ship this pull-request to prevent unexpected behavior coming from old libraries: https://github.com/meroxa/product/issues/598.

A changelog will be needed in https://github.com/meroxa/product/issues/593.

### DIFF
--- a/cmd/meroxa/turbine/golang/deploy.go
+++ b/cmd/meroxa/turbine/golang/deploy.go
@@ -24,10 +24,8 @@ func (t *turbineGoCLI) Deploy(ctx context.Context, imageName, appName, gitSha, s
 		appName,
 		"--gitsha",
 		gitSha,
-	}
-
-	if specVersion != "" {
-		args = append(args, "--spec", specVersion)
+		"--spec",
+		specVersion,
 	}
 
 	if imageName != "" {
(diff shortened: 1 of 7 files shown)


cmd/meroxa/turbine/golang/deploy.go
```diff
@@ -52,13 +50,11 @@ func (t *turbineGoCLI) Deploy(ctx context.Context, imageName, appName, gitSha, s
...
-			err = fmt.Errorf(
-				"unable to receive the deployment spec for the Meroxa Application at %s", t.appPath)
-		}
+	deploymentSpec, err = utils.GetTurbineResponseFromOutput(string(output))
+	if err != nil {
+		err = fmt.Errorf("unable to receive the deployment spec for the Meroxa Application at %s", t.appPath)
 	}
+
 	return deploymentSpec, err
 }
```
dianadoherty, line 59: I believe you can also remove this branching point:
https://github.com/meroxa/cli/blob/c4e2eb0acff3d9db3dbadbc882c9961c95e3f693/cmd/meroxa/turbine/golang/deploy.go#L119-L123

It was there before to support the old cli

dianadoherty (approved):
Looks great! Thank you for your work :)

cmd/meroxa/turbine/golang/deploy.go
raulb: @dianadoherty this is for building which is something unrelated to IR deployments. I think this is still relevant.

raulb: Update: this change was technically added unrelated to IR (https://github.com/meroxa/turbine-go/pull/88). We should still support this way of parsing turbine's library output for those running older versions of turbine.

PR #509, opened 2022-10-26
fix: Drop brackets before assessing how many functions there are in a Go app

janelletavares:
## Description of change

An empty set is interpreted as at least 1, and a function build occurs unnecessarily.

Fixes <GitHub Issue>

## Type of change

- [ ]  New feature
- [x]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

## How was this tested?

- [ ]  Unit Tests
- [x]  Tested in staging
- [ ]  Tested in minikube

## Demo

BEFORE
Screenshot from 2022-10-26 12-00-02

AFTER
Screenshot from 2022-10-26 11-58-37

### DIFF
--- a/cmd/meroxa/turbine/golang/deploy.go
+++ b/cmd/meroxa/turbine/golang/deploy.go
@@ -104,6 +104,11 @@ func (t *turbineGoCLI) NeedsToBuild(ctx context.Context, appName string) (bool, 
 	list, err := utils.GetTurbineResponseFromOutput(string(output))
 	if err == nil && list != "" {
 		// ignores any lines that are not intended to be part of the response
+		list = strings.Trim(list, "[]")
+		if list == "" {
+			return false, nil
+		}
+
 		hasFunctions := len(strings.Split(list, ",")) > 0
 		return hasFunctions, nil
 	}


cmd/meroxa/turbine/golang/deploy.go
```diff
@@ -104,6 +104,11 @@ func (t *turbineGoCLI) NeedsToBuild(ctx context.Context, appName string) (bool,
...
 	if err == nil && list != "" {
 		// ignores any lines that are not intended to be part of the response
+		list = strings.Trim(list, "[]")
+		if list == "" {
+			return false, nil
+		}
+
 		hasFunctions := len(strings.Split(list, ",")) > 0
 		return hasFunctions, nil
 	}
```
dianadoherty, line 114: I think the code below this if statement should probably also be dropped. I'm not sure when it would be used? Right now it's only if list is empty or `GetTurbineResponseFromOutput` responded with an error.

janelletavares: the code below is for older apps. we aren't requiring that people upgrade their apps, so apps that are older than a few days/a few weeks will need the regex below.  there wasn't the turbine-response tag before.

dianadoherty (approved):

Issue #535, opened 2022-11-10
[Turbine Data Flow] Add link to the Applications Details dashboard page on `meroxa apps describe`


nassor:
Expected Change: https://www.notion.so/meroxa/Turbine-Data-Flow-Implementation-Plans-6d4780375ddc4ca99a5ce7a97200bb1e#26b32621944a4384bf8df502a78ba874

janelletavares:
https://github.com/meroxa/cli/pull/538

PR #538, opened 2022-11-10
Add link to dasboard from 'apps describe' output

janelletavares:
## Description of change

Point customers to the Dashboard to visualize in more places

Fixes https://github.com/meroxa/cli/issues/535

## Type of change

- [x]  New feature
- [ ]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

## How was this tested?

- [x]  Unit Tests
- [ ]  Tested in staging
- [ ]  Tested in minikube

## Demo

| before | after |
|--------|-------|
|
Screenshot from 2022-11-10 11-45-19

|
Screenshot from 2022-11-10 11-46-22
|

### DIFF
--- a/cmd/meroxa/root/apps/describe.go
+++ b/cmd/meroxa/root/apps/describe.go
@@ -19,6 +19,7 @@ package apps
 import (
 	"context"
 	"errors"
+	"fmt"
 
 	"github.com/meroxa/cli/utils/display"
 
@@ -75,6 +76,8 @@ func (d *Describe) Execute(ctx context.Context) error {
 	d.logger.Info(ctx, output)
 	d.logger.JSON(ctx, app)
 
+	dashboardURL := fmt.Sprintf("https://dashboard.meroxa.io/apps/%s/detail", app.Name)
+	d.logger.Info(ctx, fmt.Sprintf("\n ✨ To visualize your application, visit %s", dashboardURL))
 	return nil
 }
 


cmd/meroxa/root/apps/describe.go
```diff
@@ -75,6 +76,8 @@ func (d *Describe) Execute(ctx context.Context) error {
 	d.logger.Info(ctx, output)
 	d.logger.JSON(ctx, app)
 
+	dashboardURL := fmt.Sprintf("https://dashboard.meroxa.io/apps/%s/detail", app.Name)
```
nassor, line 79: Not a review, just a comment. We have the `Name` to improve user usability, but should we use that as the default value for our URLs? The reasoning is that using UUID guarantees that the resource is unique and the same 100% of the time, and names, especially in a bigger organization, can generate start to generate conflicts/confusion.

nassor (approved):

cmd/meroxa/root/apps/describe.go
janelletavares: names are unique within the account, and you can't see the dashboard without being logged in to the account. we would have much bigger problems if a customer wanted us to change our name schema than this URL. 

    "application_name_account_id" UNIQUE, btree (name, account_id) WHERE deleted_at IS NULL
    
are you describing the situation in which a user belongs to multiple accounts and each account has an app with the same name?  that's something that the customer has control of.  i'd hope they can come up with a naming convention that doesn't confuse themselves.

jayjayjpg (approved):
This reads great, thank you for the super quick addition! :sparkles:

Issue #521, opened 2022-11-02
[Ruby: cli] Stub out Ruby cli


dianadoherty:
## Metadata 

- [Implementation plan](https://www.notion.so/meroxa/Turbine-Ruby-Implementation-Plan-ba3f655e7eea4697bf27d3fbf2fa45d3)

**Depends on**
- [x] https://github.com/meroxa/cli/issues/520

## Context 

To support Turbine Ruby, CLI has to implement a Ruby version of the [Turbine CLI interface](https://github.com/meroxa/cli/blob/a3f45174615af714697d5cf043dd8751d3bbb831/cmd/meroxa/turbine/interface.go#L7) with a Ruby CLI. 

## Acceptance 

- [ ] Stub out Ruby implementation of the [Turbine CLI interface](https://github.com/meroxa/cli/blob/a3f45174615af714697d5cf043dd8751d3bbb831/cmd/meroxa/turbine/interface.go#L7)

PR #539, opened 2022-11-10
Turbine-Ruby CLI Stub 

anna-cross:
## Description of change

Stubs for turbine-ruby cli. Contains functions needed by interfaces 

https://github.com/meroxa/cli/blob/a3f45174615af714697d5cf043dd8751d3bbb831/cmd/meroxa/turbine/interface.go#L7

Fixes https://github.com/meroxa/cli/issues/521

## Type of change

- [x]  New feature (stub)

## How was this tested?

- [x]  Unit Tests

### DIFF
--- a/cmd/meroxa/turbine/ruby/build.go
+++ b/cmd/meroxa/turbine/ruby/build.go
@@ -1,0 +1,16 @@
+package turbinerb
+
+import (
+	"context"
+)
+
+func (t *turbineRbCLI) Build(ctx context.Context, appName string, platform bool) (string, error) {
+	return "", nil
+}
+
+func (t *turbineRbCLI) CleanUpTempBuildLocation(ctx context.Context) error {
+	return nil
+}
+
+func (t *turbineRbCLI) CleanUpBinaries(ctx context.Context, appName string) {
+}
--- a/cmd/meroxa/turbine/ruby/build.go
+++ b/cmd/meroxa/turbine/ruby/build.go
@@ -2,14 +2,15 @@ package turbinerb
 
 import (
 	"context"
+	"fmt"
 )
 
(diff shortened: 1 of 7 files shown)


cmd/meroxa/turbine/ruby/run.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	"context"
+)
+
+const GrpcServerPort = 50500
+
+func (t *turbineRbCLI) Run(ctx context.Context) (err error) {
+	// Run turbine-core gRPC server
+	return err
```
dianadoherty, line 11: this will return nil, so I recommend adding an error

cmd/meroxa/turbine/ruby/build.go
```diff
@@ -0,0 +1,16 @@
+package turbinerb
+
+import (
+	"context"
+)
+
+func (t *turbineRbCLI) Build(ctx context.Context, appName string, platform bool) (string, error) {
+	return "", nil
```
dianadoherty, line 8: I think these should return a "not implemented" error

anna-cross: Sounds good!

cmd/meroxa/turbine/ruby/run.go
anna-cross: Sounds good!

dianadoherty (approved):

raulb (approved):
@dianadoherty do you think we should a feature flag for the ruby path until we finish it?

dianadoherty:
> @dianadoherty do you think we should a feature flag for the ruby path until we finish it?


because turbine ruby commands are not exposed in this PR, this PR can be merged. But we should definitely add a flag for any future pr

Issue #522, opened 2022-11-02
[Ruby: cli] Run turbine app


dianadoherty:
## Metadata 

- [Implementation plan](https://www.notion.so/meroxa/Turbine-Ruby-Implementation-Plan-ba3f655e7eea4697bf27d3fbf2fa45d3)

**Depends on**
- [x] https://github.com/meroxa/cli/issues/520
- [x] https://github.com/meroxa/cli/issues/521

## Context 

To support Turbine Ruby, CLI should be able to run a Ruby application. 

## Technical Recommendations 
1. On a `meroxa apps run`, start up a Turbine Core gRPC server in `local` mode
2. Run Turbine Ruby, and wait for a response 
3. Stop Turbine Core gRPC server 

Additional sources:
- [Example code](https://github.com/meroxa/cli/blob/d8c073883916a4e9120ed895c137aef901a2d51e/cmd/meroxa/turbine/ruby/run.go)
- [Diagram](https://www.notion.so/meroxa/Turbine-Ruby-Implementation-Plan-ba3f655e7eea4697bf27d3fbf2fa45d3#930bfcf420aa4467a30ec4a2627b43e6)

## Acceptance 

- [ ] `meroxa apps run` locally runs a Ruby app located in current directory
- [ ] `meroxa apps run --path <path>` run a Ruby app located in target directory
- [ ] `meroxa apps run` should work with and without a function

PR #540, opened 2022-11-10
Turbine-Ruby Run Command 

anna-cross:
Ruby Run Command

Run command implementation for turbine-ruby. 

Fixes  https://github.com/meroxa/cli/issues/522

## Type of change

- [x]  New feature

## How was this tested?

- [x]  Unit Tests
- [x]  Tested in minikube

To test: 

1. Clone https://github.com/meroxa/turbine-examples/tree/main/ruby/simple
2. Build this cli branch 
3. Do any necessary ruby installs, follow : https://www.notion.so/meroxa/Turbine-Ruby-Installation-Guide-Testing-Cases-1317138171674e1b8fe5acad5705214e
4. Run `meroxa app run` in the simple app dir 

## Demo

(ignore the warnings, its just my M1 issues) 

App run with function 

image

App run without function 

image

App run with path flag 

image


cmd/meroxa/root/apps/apps.go
```diff
@@ -24,7 +24,7 @@ import (
 
 type Apps struct{}
 
-const LanguageNotSupportedError = "Currently, we support \"javascript\", \"golang\", and \"python\""
+const LanguageNotSupportedError = "Currently, we support \"javascript\", \"golang\", \"python\", and \"ruby (beta)\" "
```
raulb, line 27: Not part of this pull-request, but I think we should start centralizing language support in turbine-core so we remove these from the CLI. I have a TODO item for this, so I'll make sure it gets done.

raulb (approved):
Looks great, nice work.

cmd/meroxa/root/apps/apps.go
anna-cross: Sounds good! Thanks Raul!

PR #543, opened 2022-11-14
Add version function for each turbine library

janelletavares:
## Description of change

Determine the version of turbine-go used in an App
make grabbing all versions of turbine libs used in an App accessible

Fixes https://github.com/meroxa/turbine-project/issues/266

## Type of change

- [x]  New feature
- [ ]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

## How was this tested?

- [ ]  Unit Tests
- [ ]  Tested in staging
- [ ]  Tested in minikube

## Demo

| before | after |
|--------|-------|
|||

### DIFF
--- a/cmd/meroxa/turbine/golang/version.go
+++ b/cmd/meroxa/turbine/golang/version.go
@@ -29,8 +29,9 @@ func (t *turbineGoCLI) GetVersion(ctx context.Context) (string, error) {
 	chars := []rune(version)
 	if chars[0] == 'v' {
 		// Looks like v0.0.0-20221024132549-e6470e58b719
+		const sections = 3
 		parts := strings.Split(version, "-")
-		if len(parts) < 3 {
+		if len(parts) < sections {
 			return "", fmtErr
 		}
 		version = parts[2]
--- a/cmd/meroxa/turbine/golang/version.go
+++ b/cmd/meroxa/turbine/golang/version.go
@@ -17,7 +17,7 @@ func (t *turbineGoCLI) GetVersion(ctx context.Context) (string, error) {
 		"list", "-m", "-f", "'{{ .Version }}'", "github.com/meroxa/turbine-go")
 	cmd.Dir = t.appPath
 	fmtErr := fmt.Errorf(
-		"unable to determine the version of turbine-js-cli used for the Meroxa Application at %s",
+		"unable to determine the version of turbine-go used by the Meroxa Application at %s",
 		t.appPath)
 
 	stdout, err := cmd.CombinedOutput()
--- a/cmd/meroxa/turbine/golang/version.go
+++ b/cmd/meroxa/turbine/golang/version.go
@@ -1,0 +1,39 @@
+package turbinego
+
+import (
+	"context"
+	"fmt"
+	"os/exec"
+	"strings"
+)
+
+// GetVersion will return the tag or hash of the turbine-go dependency of a given app.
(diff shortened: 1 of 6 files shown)


cmd/meroxa/turbine/golang/version.go
```diff
@@ -0,0 +1,40 @@
...
+func (t *turbineGoCLI) GetVersion(ctx context.Context) (string, error) {
+	var cmd *exec.Cmd
+
+	cmd = exec.CommandContext(
+		ctx,
+		"go",
+		"list", "-m", "-f", "'{{ .Version }}'", "github.com/meroxa/turbine-go")
+	cmd.Dir = t.appPath
+	fmtErr := fmt.Errorf(
+		"unable to determine the version of turbine-js-cli used for the Meroxa Application at %s",
```
nassor, line 20: turbine-js-cli -> turbine-go

cmd/meroxa/turbine/javascript/version.go
```diff
@@ -0,0 +1,26 @@
...
+
+	utils "github.com/meroxa/cli/cmd/meroxa/turbine"
+)
+
+// GetVersion calls turbine-js-cli to get the version of the app.
+func (t *turbineJsCLI) GetVersion(ctx context.Context) (string, error) {
+	cmd := utils.RunTurbineJS(ctx, "version", t.appPath)
+	stdOut, err := utils.RunCmdWithErrorDetection(ctx, cmd, t.logger)
+	fmtErr := fmt.Errorf(
+		"unable to determine the version of turbine-js-cli used for the Meroxa Application at %s",
```
nassor, line 15: turbine-js-cli -> turbine-js

cmd/meroxa/turbine/golang/version.go
janelletavares:
```suggestion
		"unable to determine the version of turbine-go used by the Meroxa Application at %s",
```

cmd/meroxa/turbine/javascript/version.go
janelletavares: there are two different packages now... it's not just turbine-js.  it might actually be turbine-js-framework version... i'll circle back

nassor (approved):

jayjayjpg (approved):
This reads great, thank you for adding this feature so swiftly! :sparkles:

cmd/meroxa/turbine/javascript/version.go
janelletavares: @jayjayjpg do you know what's most accurate here?

jayjayjpg: @janelletavares The package itself is called `turbine-js-cli` (see also [here](https://github.com/meroxa/turbine-js/tree/master/packages/turbine-js-cli)), but I'll leave the decision on what the best wording is up to you!

The advantage of using the more specific `turbine-js-cli` wording is that a user would have a very specific error message and if they wanted to dive into the source code to see what's happening they could do so more easily. The advantage of using the `turbine-js` wording would be that we don't expose the customer to library-internal concepts that they may never have to interact with anyways

janelletavares: i don't think the separation of the packages is only internal.  we specify which packages we're manipulating in logs of the upgrade command.  which customers don't have to use at all... they can manage the package.json in their app directly, which is what most of javascript developers do, right?

PR #546, opened 2022-11-15
Add a Referer header for Usage Metrics

janelletavares:
## Description of change

Indicate clearly that a request came from the CLI from the Platform API for Usage Metrics

Fixes https://github.com/meroxa/platform-api/issues/1525

## Type of change

- [x]  New feature
- [ ]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

## How was this tested?

- [ ]  Unit Tests
- [ ]  Tested in staging
- [ ]  Tested in minikube

## Demo

| before | after |
|--------|-------|
|||

### DIFF
--- a/vendor/github.com/meroxa/meroxa-go/pkg/meroxa/options.go
+++ b/vendor/github.com/meroxa/meroxa-go/pkg/meroxa/options.go
@@ -80,13 +80,24 @@ func WithAuthentication(conf *oauth2.Config, accessToken, refreshToken string, o
 	}
 }
 
-// WithClient sets the http client to use for requests.
+// WithAccountUUID sets the http client to use for requests.
 func WithAccountUUID(accountUUID string) Option {
 	return func(r *Requester) error {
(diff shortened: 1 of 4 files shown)


vendor/github.com/meroxa/meroxa-go/pkg/meroxa/options.go
```diff
@@ -90,3 +90,14 @@ func WithAccountUUID(accountUUID string) Option {
...
 	}
 }
+
+// WithReferer sets the Referer header to use for requests.
+func WithReferer(referer string) Option {
+	return func(r *Requester) error {
+		if r.headers == nil {
+			r.headers = make(http.Header)
+		}
+		r.headers["Referer"] = []string{referer}
```
nassor, line 100: The Referer header always expects a URL; we shouldn't use it for something else.

Reference: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referer

PR #567, opened 2022-12-07
feat: Update Python Upgrade Logic

ericcheatham:
## Description of change

Fixes upgrade logic path for turbine-py applications. Makes the following assumptions/actions:
- Runs `pip upgrade` regardless of what version of turbine py is present (most recent or not). This is more in line with other language logic paths
- Pulls the version to add to requirements text file from turbine-py itself which should be the most recent version 
- Attempts to write updated verion to requirements text file 
  -  Wondering if it's better to prompt a user to make that change themselves? 

## Type of change

- [ ]  New feature
- [x]  Bug fix
- [x]  Refactor
- [ ]  Documentation

## How was this tested?

- [ ]  Unit Tests
- [ ]  Tested in staging
- [ ]  Tested in minikube

## Demo

| before | after |
|--------|-------|
|||

### DIFF
--- a/cmd/meroxa/turbine/python/upgrade.go
+++ b/cmd/meroxa/turbine/python/upgrade.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os/exec"
 
+	utils "github.com/meroxa/cli/cmd/meroxa/turbine"
 	"github.com/meroxa/cli/log"
 )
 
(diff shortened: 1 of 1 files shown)


cmd/meroxa/turbine/python/upgrade.go
```diff
@@ -4,41 +4,41 @@ import (
...
 
+	utils "github.com/meroxa/cli/cmd/meroxa/turbine"
 	"github.com/meroxa/cli/log"
 )
 
 const turbinePYVersion = "1.5.3"
 
 // Upgrade fetches the latest Meroxa dependencies.
 func (t *turbinePyCLI) Upgrade(vendor bool) error {
-	cmd := exec.Command("grep", "turbine-py==", "requirements.txt")
```
ericcheatham, line 14: We're choosing to run `pip upgrade` specifically on our libraries regardless of what is installed. Nothing would change if they are on the most up to date version.

raulb:
@ericcheatham going through open PRs. Do you think this is something you'll get to or it's safe to close it for now? Let me know, thank you!

Issue #526, opened 2022-11-03
Git init settings do not work with older versions of git


janelletavares:
**Describe the bug**  

https://stackoverflow.com/questions/42871542/how-can-i-create-a-git-repository-with-the-default-branch-name-other-than-maste

With git version 2.25, the init.defaultBranch setting was not taking.  Differentiate between before or after 2.28.

**Machine (please complete the following information):**  

 - CLI Version [e.g. 22]

**To Reproduce**  

What can we do to reproduce your bug?

Please provide:
- Relevent CLI commands
- Relevent environment details

**Expected behavior**  
What did you expect to happen? 

**Screenshots**  
If applicable, add screenshots to help explain your problem. (Be sure to scrub any sensitive information.)

**Additional context**

janelletavares:
https://github.com/meroxa/cli/pull/569

PR #569, opened 2022-12-07
fix: git init with older versions of git

janelletavares:
## Description of change

Setting the default branch is different depending on the current version of git

Fixes https://github.com/meroxa/cli/issues/526

## Type of change

- [ ]  New feature
- [ ]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

## How was this tested?

- [ ]  Unit Tests
- [ ]  Tested in staging
- [ ]  Tested in minikube

## Demo

| before | after |
|--------|-------|
|||


cmd/meroxa/turbine/utils_test.go
```diff
@@ -39,3 +40,8 @@ func TestGitInit(t *testing.T) {
 
 	os.RemoveAll(testDir)
 }
+
+func TestCheckGitVersion(t *testing.T) {
+	val := checkGitVersion(context.Background())
+	assert.False(t, val)
```
jayjayjpg, line 46: I don't know a lot about Go and I'm curious to learn more in this section: what does `assert.False(t, val)` do? I assume this test will only pass if `t` is empty or false, but I don't fully understand yet, how this will check for the `checkGitVersion` to return the correct value. Or does `assert.False()` compare `t` and `val` with each other instead?

jayjayjpg (approved):
This reads great and thank you for tackling this! I added one more question to understand the test coverage better, but this is more out of curiosity, rather than a suggestion for a change!

cmd/meroxa/turbine/utils.go
```diff
@@ -182,18 +184,43 @@ func GitInit(ctx context.Context, appPath string) error {
...
+
+	if oldWay {
+		cmd := exec.CommandContext(ctx, "git", "checkout", "-b", "main")
+		cmd.Path = appPath
+		_ = cmd.Run()
+	}
 	return nil
 }
 
+func checkGitVersion(ctx context.Context) bool {
```
raulb, line 209: I think we should return error on `cmd.CombinedOutput`. Aside, I wonder we could make this a bit clearer by making these tweaks.

@janelletavares what do you think?

```suggestion
func checkGitVersion(ctx context.Context) (bool, error) {
	cmd := exec.CommandContext(ctx, "git", "version")
	out, err := cmd.CombinedOutput()
	if err != nil {
	  return false, err
	}

	r := regexp.MustCompile("git version ([0-9.]+)")
	matches := r.FindStringSubmatch(string(out))
	if len(matches) > 0 {
		comparison := semver.Compare("2.28", matches[1])
		return comparison >= 1
	}
	return true
}
```

cmd/meroxa/turbine/utils.go
```diff
@@ -182,18 +184,43 @@ func GitInit(ctx context.Context, appPath string) error {
 		return errors.New("path is required")
 	}
 
-	cmd := exec.Command("git", "config", "--global", "init.defaultBranch", "main")
-	cmd.Path = appPath
-	_ = cmd.Run()
+	oldWay := checkGitVersion(ctx)
```
raulb, line 187: I think `oldWay` is not explanatory enough to know what it does unless checking the `checkGitVersion` function. What do you think we change it to something like this?:

I also made some changes to ensure we return err if it's found when doing `cmd.Run`.

```go
func GitInit(ctx context.Context, appPath string) error {
	if appPath == "" {
		return errors.New("path is required")
	}
      
       isGitOlderThan228 := checkGitVersion(ctx)

	if !isGitOlderThan228 {
		cmd := exec.CommandContext(ctx, "git", "config", "--global", "init.defaultBranch", "main")
		cmd.Path = appPath
		err := cmd.Run()
                 if err != nil {
                      return err
                 }
	}

	cmd := exec.CommandContext(ctx, "git", "init", appPath)
	output, err := cmd.CombinedOutput()
	if err != nil {
		return fmt.Errorf(string(output))
	}

	if isGitOlderThan228 {
		cmd := exec.CommandContext(ctx, "git", "checkout", "-b", "main")
		cmd.Path = appPath
		_ = cmd.Run()
	}
	return nil
}
```

raulb (changes requested):
Made some suggestions to make the code a bit clearer in my opinion. 

@janelletavares let me know what you think.

cmd/meroxa/turbine/utils_test.go
janelletavares: t is a pointer to the whole unit testing framework for Golang
we are asserting that val is always false here.  I wanted this code to at least run once in unit tests, but it can only do one thing because the environmental version of git is the environmental version of git, so there's just this one test case.  val == false here means that the version of git is greater than 2.28. the current release is 2.39, so github should be running something close to that.

cmd/meroxa/turbine/utils.go
janelletavares: happy to change the returns

i mentioned above that i wanted to error here as little as possible and just have some kind of sane default behavior.  i could go either way.

cmd/meroxa/turbine/utils.go
janelletavares: looking at the workflow again... this is the last step in init.  if the customer was savvy enough, they could figure out what was left to do and do it themselves. it's not too annoying. i'll update.

raulb (approved):

cmd/meroxa/turbine/utils_test.go
jayjayjpg: I see, that sounds great and thank you for the outline! 🙇🏾‍♀️

Issue #594, opened 2023-01-20
[Env: cli] Environment support for `apps ls`


dianadoherty:
## Metadata 

- [Implementation plan](https://www.notion.so/meroxa/Environments-Turbine-Support-Implementation-Plan-8aa02e78eaec453ab6b47ff9d5c2a9ed)

**Depends on**

- [x] https://github.com/meroxa/meroxa-go/issues/153

## Context 

Ensure that `meroxa apps ls` reports an App's Environment information

## Acceptance 

- [x]  Add an Environment name column when listing Environments with: `meroxa apps ls` 
    - Note: a column doesn't need to be reported if customer's don't have the Environments feature enabled

raulb:
@dianadoherty rather than relying on feature flag, I'm going to rely on the information that's stored. i.e.: if a user doesn't have the environments feature flag enabled, applications shouldn't have the environment associated. If it does (and different than common), I think it's worth showing the information. Let me know if that works for you.

PR #603, opened 2023-01-27
feat(apps): Include environment info in `apps ls`

raulb:
## Description of change

Fixes https://github.com/meroxa/cli/issues/594

## Type of change

- [x]  New feature
- [ ]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

## How was this tested?

- [x]  Unit Tests
- [ ]  Tested in staging
- [ ]  Tested in minikube

## Demo

**Before**

```
UUID      NAME           LANGUAGE   GIT SHA   STATE   
====== ================ ========== ========= ======= 
8ed... application-name golang     ad87...   running   
```

**Now with no environment Info, or if environment is common**

Same as _**Before**_

**With environment Info (that's not common)**

```
UUID      NAME           LANGUAGE   GIT SHA   STATE   ENVIRONMENT 
====== ================ ========== ========= ======= =============
8ed... application-name golang     ad87...   running   my-env   
```

### DIFF
--- a/cmd/meroxa/root/apps/list_test.go
+++ b/cmd/meroxa/root/apps/list_test.go
@@ -23,10 +23,12 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/meroxa/cli/utils"
+
+	"github.com/meroxa/cli/log"
+
 	"github.com/golang/mock/gomock"
 
-	"github.com/meroxa/cli/cmd/meroxa/turbine"
-	"github.com/meroxa/cli/log"
 	"github.com/meroxa/cli/utils/display"
 	"github.com/meroxa/meroxa-go/pkg/meroxa"
 	"github.com/meroxa/meroxa-go/pkg/mock"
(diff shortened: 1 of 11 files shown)


dianadoherty (approved):
My one concern on not relying on feature flags to determine whether we should display env info or not is for customers who might have an app in an env, and another in a common env. That customer will have an inconsistent view where they sometimes see envs, and other times they don't. 

However, looking at how we decided to display connector env info, it seems like the same approach to not rely on feature flags was taken. So I'm good to go with the current approach

raulb:
> My one concern on not relying on feature flags to determine whether we should display env info or not is for customers who might have an app in an env, and another in a common env. That customer will have an inconsistent view where they sometimes see envs, and other times they don't.

@dianadoherty good point. I think it's however a tradeoff we can take. If this becomes confusing I could add the check for feature flags. However, I think if some information is not present (e.g.: an application without an environment), it's a decent tradeoff not showing anything. If we decide to backfill applications with a common environment, then the information would show up in there so I think it's fine.

PR #617, opened 2023-02-01
feat: make callback url configurable

raulb:
## Description of change

In order to span up a new browser in a different environment (such as the one using Nimbus), we'll need to make callback urls configurable. This pull request should take care of that.

## Type of change

- [x]  New feature
- [ ]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

## How was this tested?

⚠️ Will test on a nimbus environment as soon as I have one available.

- [ ]  Tested in nimbus.

## Demo

**Before**

Callback url is always "http://localhost:21900/oauth/callback"

**Now**

You can set different environment variables that would make it configurable:

- `MEROXA_AUTH_CALLBACK_HOST`
- `MEROXA_AUTH_CALLBACK_URL`
- `MEROXA_AUTH_CALLBACK_PORT`
- `MEROXA_AUTH_CALLBACK_PROTOCOL`

⚠️ `MEROXA_AUTH_CALLBACK_URL` will take precedence over the rest in the event of that being set.

Or simply do (for example):

`MEROXA_AUTH_CALLBACK_URL=http://nimbus.meroxa.io/oauth/callback meroxa login`


simonl2002 (approved):
lgtm

PR #619, opened 2023-02-01
Ensure turbine resource is in the same environment as the app

### LINKED ISSUE #593: [Env: cli] Environment apps validations
## Metadata 

- [Implementation plan](https://www.notion.so/meroxa/Environments-Turbine-Support-Implementation-Plan-8aa02e78eaec453ab6b47ff9d5c2a9ed)

## Context 

We want to ensure that validations to create an Application within a specific Environment are properly configured. 

## Acceptance 

- [x] Validation: if an Application exists in an Environment, prevent another deploy to a separate Environment
    - Note: exclude this validation for feature branches
    - Note: this feature might already exist. If it does, ensure it works as expected
- [x] Validation: [when validating if resources mentioned in the Application exist](https://github.com/meroxa/cli/blob/2beca0629ecac6cd32e101da26b67b50676c2df9/cmd/meroxa/root/apps/deploy.go#L565-L587), ensure that resource exists in the Environment the app is getting created in

lyuboxa:
## Description of change

Turbine apps provisioned on env only supports resource which are part of these environments.
This change adds validation to ensure this is the case.

Minor refactor, rename method and ensure it returns errors rather than strings.
Create file to store error messages, it is a start.
Flatten bunch of obvious conditional paths.

Fixes #593

## Type of change

- [x]  New feature
- [ ]  Bug fix
- [x]  Refactor
- [ ]  Documentation

## How was this tested?

- [x]  Unit Tests
- [ ]  Tested in staging
- [ ]  Tested in minikube

## Demo

| before | after |
|--------|-------|
|||

### DIFF
--- a/cmd/meroxa/root/apps/deploy.go
+++ b/cmd/meroxa/root/apps/deploy.go
@@ -583,12 +583,21 @@ func (d *Deploy) validateResources(ctx context.Context, rr []turbine.Application
 			errs = append(errs, wrapNotFound(r.Name, err))
 		case resource.Status.State != meroxa.ResourceStateReady:
 			errs = append(errs, fmt.Errorf("resource %q is not ready and usable", r.Name))
+		// app is provisioned in common env, but resource was added at self hosted env
+		case d.flags.Environment == "" && resource.Environment != nil:
+			errs = append(errs, fmt.Errorf(
+				"resource %q is in common env, but app is in %q",
+				r.Name,
+				resource.Environment.Name,
+			))
+		// app is provisioned in an env, but resource is in common env
 		case d.flags.Environment != "" && resource.Environment == nil:
 			errs = append(errs, fmt.Errorf(
 				"resource %q is not in app env %q, but in common",
 				r.Name,
 				d.flags.Environment,
 			))
+		// app is provisioned in an env, but resource is in different self hosted env
 		case d.flags.Environment != "" && resource.Environment.Name != d.flags.Environment:
 			errs = append(errs, fmt.Errorf(
 				"resource %q is not in app env %q, but in %q",
@@ -631,6 +640,7 @@ func (d *Deploy) checkResourceAvailability(ctx context.Context) error {
 		return err
 	}
 
+	d.logger.StopSpinnerWithStatus("Can access your Turbine resources", log.Successful)
 	return nil
 }
 
--- a/cmd/meroxa/root/apps/deploy.go
+++ b/cmd/meroxa/root/apps/deploy.go
@@ -567,6 +567,7 @@ func (d *Deploy) validateAppJSON(ctx context.Context) error {
 
 func (d *Deploy) validateResources(ctx context.Context, rr []turbine.ApplicationResource) error {
 	var errs []error
+	validated := make(map[string]bool)
 
 	wrapNotFound := func(name string, err error) error {
 		if strings.Contains(err.Error(), "could not find") {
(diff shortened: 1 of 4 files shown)


cmd/meroxa/root/apps/errors.go
```diff
@@ -0,0 +1,6 @@
+package apps
+
+const (
+	resourceInvalidError = "⚠️  Run 'meroxa resources list' to verify that the resource names " +
```
raulb, line 4: I like this approach. We could expand it from here.

cmd/meroxa/root/apps/deploy.go
```diff
@@ -535,56 +537,69 @@ func (d *Deploy) validateAppJSON(ctx context.Context) error {
...
-			return err
-		}
-		config, err = turbine.ReadConfigFile(d.path)
-		d.appConfig = &config
-		if err != nil {
-			return err
-		}
+	// exit early if app config is loaded
+	if d.appConfig != nil {
+		return d.validateLanguage()
```
raulb, line 542: We need to validate language after `GetLangFromAppJSON` since that's where we assign language.

raulb (changes requested):
@lyuboxa I think everything looks good except this pr is changing the order of language assignment and validation so we need to make sure that's still correct. Let me know if you'd need to pair on this (kind of messy) code.

cmd/meroxa/root/apps/deploy.go
lyuboxa: This only happens if `appConfig` is nil, as per https://github.com/meroxa/cli/blob/eae9dc253cf508b008b028e11db87b3ea1276be4/cmd/meroxa/root/apps/deploy.go#L538-L557
Was the original code wrong?

raulb: @lyuboxa I see it. Yeah, I think it's fine the way it is now.

raulb (approved):
Approving. Sounds like @lyuboxa will do manual testing just in case .

cmd/meroxa/root/apps/deploy.go
```diff
@@ -601,21 +616,21 @@ func (d *Deploy) checkResourceAvailability(ctx context.Context) error {
...
+		d.logger.StopSpinnerWithStatus("Can access your Turbine resources", log.Successful)
+		return nil
+	}
+
+	if err := d.validateCollections(ctx, resources); err != nil {
+		d.logger.StopSpinnerWithStatus("Resource availability check failed", log.Failed)
+		return err
 	}
 
-	d.logger.StopSpinnerWithStatus("Can access your Turbine resources", log.Successful)
```
dianadoherty, line 618: You'll need to return this log line. Otherwise the cli goes crazy

dianadoherty (changes requested):

cmd/meroxa/root/apps/deploy.go
```diff
@@ -535,56 +537,69 @@ func (d *Deploy) validateAppJSON(ctx context.Context) error {
...
+				d.flags.Environment,
+			))
+		case d.flags.Environment != "" && resource.Environment.Name != d.flags.Environment:
+			errs = append(errs, fmt.Errorf(
+				"resource %q is not in app env %q, but in %q",
+				r.Name,
+				d.flags.Environment,
+				resource.Environment.Name,
+			))
+		}
```
dianadoherty, line 599: One more missing case is if app is getting deployed in the common env, but the resource is in another env

cmd/meroxa/root/apps/deploy.go
lyuboxa: uh oh you are right!  thanks

cmd/meroxa/root/apps/deploy.go
lyuboxa: Thanks will add it, good catch

lyuboxa: fixed in 798a4c17

cmd/meroxa/root/apps/deploy.go
lyuboxa: fixed in 798a4c17

dianadoherty (approved):

Issue #622, opened 2023-02-03
[Env: cli] Apps ls and describe should not expect env type and provider


anna-cross:
## Context 

Only pull  env name and uuid into ls and description, remove provider and type from the table 

## Acceptance 

- [ ]  On app ls and app describe, show env uuid and name

PR #625, opened 2023-02-03
Env app describe and ls changes 

anna-cross:
## Description of change

We are no longer pulling type and provider info for Envs on Apps. Making changes according to that. 

Fixes <https://github.com/meroxa/cli/issues/622>

## Type of change

- [x]  Refactor

## How was this tested?

- [x]  Unit Tests
- [x]  Tested in staging
- [ ]  Tested in minikube

## DEMO 

# APP DESCRIBE WITH PRIVATE ENV AND  IN COMMON 

image

# APP LS WITH PRIVATE ENV AND IN COMMON 

image

# APP LS JUST COMMON 

image

# APP LS JUST PRIVATE ENV 

image

### DIFF
--- a/utils/display/apps.go
+++ b/utils/display/apps.go
@@ -18,8 +18,6 @@ func AppsTable(apps []*meroxa.Application, hideHeaders bool) string {
 		return ""
 	}
 
-	withEnvironment := false
-
 	table := simpletable.New()
 
 	for _, app := range apps {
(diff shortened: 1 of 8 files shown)


utils/display/apps.go
```diff
@@ -37,10 +33,13 @@ func AppsTable(apps []*meroxa.Application, hideHeaders bool) string {
...
 		env, err := app.Environment.GetNameOrUUID()
 
-		if err == nil && itShouldEnvInfo(app.Environment) {
+		if err == nil {
 			withEnvironment = true
 			r = append(r, &simpletable.Cell{Align: simpletable.AlignLeft, Text: env})
 		}
+		if withEnvironment == true && err != nil {
+			r = append(r, &simpletable.Cell{Align: simpletable.AlignLeft, Text: "common"})
+		}
```
dianadoherty, line 42: Logic here seems a little weird
I feel like we can also check if env is nil with `app.Environment != nil` for whether the app is on a common env vs not

anna-cross: So, in this I'm trying to cover a case  when we do add an Environment column on table for an app with a private env, but there is also an app without environment (aka common). 
If app is common, app.Environment will be nil , but we would still want to add a text as the simple table starts panicking when Environment column exists but we pass it no value in a row 🤔

dianadoherty: Yeah, but the only time `withEnvironment` is set to true if `err == nil`, so the second if statement will never work

anna-cross: Looking at it again, ye it makes no sense. Made some changes. We are now adding Env column no matter what and just filling it with common if an app doesn't have a private app

utils/display/apps.go
```diff
@@ -35,12 +29,15 @@ func AppsTable(apps []*meroxa.Application, hideHeaders bool) string {
...
-		env, err := app.Environment.GetNameOrUUID()
-
-		if err == nil && itShouldEnvInfo(app.Environment) {
-			withEnvironment = true
-			r = append(r, &simpletable.Cell{Align: simpletable.AlignLeft, Text: env})
+		if app.Environment != nil {
+			env, err := app.Environment.GetNameOrUUID()
+			if err == nil {
+				r = append(r, &simpletable.Cell{Align: simpletable.AlignLeft, Text: env})
+			}
```
dianadoherty, line 36: The error case isn't getting handled here. Instead I recommend being explicit about whether you want to include the name or the uuid here. 

Here's an example for resources https://github.com/meroxa/cli/blob/2a0ff49ec84721823307244176272a18b0bd5437/utils/display/resources.go#L53-L65

dianadoherty (approved):

Issue #623, opened 2023-02-03
[Env: cli] Remove support for "common" as an env flag input


anna-cross:
## Context 

We want to remove the ability of user to pass "common" as an environment flag across all resources. This creates confusion as you cannot have a private env named common and in the background in cli we default to common env when they pass it as a flag. 

## Acceptance 

- [ ]  User cannot pass "common" as an env flag

PR #626, opened 2023-02-03
Remove cli logic to handle common env for api

dianadoherty:
## Description of change

Remove cli logic to handle common env for api

Fixes https://github.com/meroxa/cli/issues/623

## Type of change

- [x]  New feature
- [ ]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

## How was this tested?

- [ ]  Unit Tests
- [x]  Tested in staging
- [ ]  Tested in minikube

### Before
```
meroxa resource create pg --type postgres --url  $PG_URL --env common 
Creating "postgres" resource in "common" environment...
Resource "pg" is successfully created!
```

### After 

```
meroxa resource create pg-c --type postgres --url $PG_URL --env common
Creating "postgres" resource in "common" environment...
Error: could not find environment. 1 detail reported:
1. type: "postgres"

```

### DIFF
--- a/cmd/meroxa/root/pipelines/create.go
+++ b/cmd/meroxa/root/pipelines/create.go
@@ -76,8 +76,7 @@ func (c *Create) Execute(ctx context.Context) error {
 		p.Metadata = metadata
 	}
 
-	// If the environment specified is not the common environment.
-	if c.flags.Environment != "" && c.flags.Environment != string(meroxa.EnvironmentTypeCommon) {
+	if c.flags.Environment != "" {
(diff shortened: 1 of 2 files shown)


raulb (approved):
I like this change. I'd rather not being clever and assume that's the user intend and really let the Platform API handle the default environment which should be common. This will let us change the Platform API without risking a client like CLI doing something we might not expect. 

I also think this is much better for the user. If environment is not specified, then we should assume it's "common".

Thank you for making this change @dianadoherty

PR #654, opened 2023-03-17
Add apps open command

janelletavares:
## Description of change

Convenience methods for seeing app details in the dashboard and interacting with it there. 

Fixes https://github.com/meroxa/turbine-project/issues/219

## Type of change

- [x]  New feature
- [ ]  Bug fix
- [ ]  Refactor
- [x]  Documentation

## How was this tested?

- [x]  Unit Tests
- [ ]  Tested in staging
- [ ]  Tested in minikube

## Demo

| before | after |
|--------|-------|
|||


cmd/meroxa/root/apps/open.go
```diff
@@ -0,0 +1,107 @@
...
+meroxa apps open --path /my/app
+meroxa apps open NAMEorUUID`,
+		Beta: true,
+	}
+}
+
+func (o *Open) Execute(ctx context.Context) error {
+	nameOrUUID := o.args.NameOrUUID
+	if nameOrUUID != "" && o.flags.Path != "" {
+		return fmt.Errorf("supply either NamrOrUUID argument or path flag")
```
jayjayjpg, line 79: I think this would be:

```suggestion
		return fmt.Errorf("supply either NameOrUUID argument or --path flag")
```

cmd/meroxa/root/apps/open.go
```diff
@@ -0,0 +1,107 @@
...
+		}
+		nameOrUUID = config.Name
+	}
+
+	// open a browser window to the application details
+	dashboardURL := fmt.Sprintf("https://dashboard.meroxa.io/apps/%s/detail", nameOrUUID)
+	err := open.Start(dashboardURL)
+	if err != nil {
+		o.logger.Errorf(ctx, "meroxa: can't open browser to URL %s: %s\n", dashboardURL, err)
+		//os.Exit(1)
```
jayjayjpg, line 100: Would this one be included or better be removed?

cmd/meroxa/root/apps/open_test.go
jayjayjpg, line 1: NIce test coverage! :sparkles:

jayjayjpg (commented):
This reads great so far and I can also give this a try locally in just a bit!

cmd/meroxa/root/apps/open.go
janelletavares: thanks!

cmd/meroxa/root/apps/open.go
```diff
@@ -0,0 +1,107 @@
...
+		config, err := turbine.ReadConfigFile(o.path)
+		if err != nil {
+			return err
+		}
+		nameOrUUID = config.Name
+	}
+
+	// open a browser window to the application details
+	dashboardURL := fmt.Sprintf("https://dashboard.meroxa.io/apps/%s/detail", nameOrUUID)
+	err := open.Start(dashboardURL)
```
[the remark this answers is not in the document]
janelletavares: then i'd have to return nil on 102 because err wouldn't be defined. either way

cmd/meroxa/root/apps/open.go
janelletavares: oh, i meant to check that this was necessary from the code that i copied... it's hard to test because the binary being used is compiled in... oh well, i'll just remove it.

cmd/meroxa/root/apps/open.go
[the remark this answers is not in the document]
janelletavares: actually, i want to return the error and not just print it for proper exit codes.  so then maybe i don't want to print it?

cmd/meroxa/root/apps/open_test.go
```diff
@@ -0,0 +1,170 @@
...
+		path         string
+		appArg       string
+		appFlag      string
+		errSubstring string
+	}{
+		{
+			desc:   "Successfully open app link with arg",
+			appArg: "app-name",
+			path: func() string {
+				return filepath.Join("/tmp", uuid.New().String()) //nolint:gocritic
```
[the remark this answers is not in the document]
janelletavares: i want it to be a different path for each use case, so i don't want to reuse this path value, but not it doesn't need to be a function. i copied the nolint from somewhere else, but it seems like it doesn't need to be there at all!

janelletavares:
> This already works great! One thing I noticed, that I'm not sure about if we should solve it: `meroxa apps open` currently only works in production application but wouldn't be something that we could use in other test environments, such as `staging`:
> 
@jayjayjpg yeah, I noticed that too.  Everywhere that we mention the dashboard in the CLI, the URL is hardcoded to the prod version, so i think we're stuck with this for now, and we'd have to change something across the board (maybe checking one of the environment variables, which would be a small ticket) to have testing support for staging URLs.

cmd/meroxa/root/apps/open.go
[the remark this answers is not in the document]
janelletavares:
```
$ m apps open --path ../demos/tenth
can't open browser to URL https://dashboard.meroxa.io/apps/tenth/detail
Error: forcing
```
looks like this now

jayjayjpg:
> @jayjayjpg yeah, I noticed that too. Everywhere that we mention the dashboard in the CLI, the URL is hardcoded to the prod version, so i think we're stuck with this for now, or we'd have to change something across the board (maybe checking one of the environment variables, which would be a small ticket) to have testing support for staging URLs.

@janelletavares Sounds good and thank you for confirming!

PR #660, opened 2023-03-21
Apps and build commands are no longer beta

janelletavares:
## Description of change

Removing the beta tag

Fixes <GitHub Issue>

## Type of change

- [ ]  New feature
- [ ]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

## How was this tested?

- [ ]  Unit Tests
- [ ]  Tested in staging
- [ ]  Tested in minikube

## Demo

| before | after |
|--------|-------|
|||


Makefile
```diff
@@ -37,6 +37,7 @@ fig:
 
 .PHONY: lint
 lint:
+	docker pull golangci/golangci-lint:latest
```
jayjayjpg, line 40: Is this needed to ensure we're using a newer version of the linter moving forward?

cmd/meroxa/root/apps/deploy.go
```diff
@@ -498,7 +497,7 @@ func (d *Deploy) getAppImage(ctx context.Context) (string, error) {
 	d.localDeploy.Lang = d.lang
 	d.localDeploy.AppName = d.appName
 	if d.localDeploy.Enabled {
-		fqImageName, err = d.localDeploy.GetDockerImageName(ctx, d.logger, d.path, d.appName, d.lang)
+		fqImageName, err = d.localDeploy.GetDockerImageName(ctx, d.logger, d.path, d.appName)
```
jayjayjpg, line 500: Has usage of the `d.lang` parameter been removed from the `GetDockerImageName` function in the past and this is why it can be omitted from now on? Or is it better to revert this change?

cmd/meroxa/root/apps/list_test.go
```diff
@@ -55,7 +55,7 @@ func TestListAppsExecution(t *testing.T) {
 		{
 			desc: "With applications in a private and common environment",
 			apps: func() []*meroxa.Application {
-				aa := utils.GenerateApplicationWithEnv("", meroxa.EnvironmentTypePrivate, meroxa.EnvironmentProviderAws)
+				aa := utils.GenerateApplicationWithEnv("")
```
jayjayjpg, line 58: Is the removal of the `meroxa.Environment*` parameters needed?

jayjayjpg (commented):
This reads great so far, but I also noticed that this PR also includes a couple of other changes with regards to environments work that seems to be unrelated to the removal of the beta tag -- would it be good to rebase the branch onto the latest `master` and pick out the beta changes?

Let me know if I'm missing something and happy to take another look!

cmd/meroxa/root/apps/deploy.go
janelletavares: it is currently not used, and i'm not sure off the top of my head how it was ever used.  the linter seems to have released a new version that has a whole lot more complaints to deal with, mostly related to unused arguments.  this PR would not pass the linter check without making these changes.  my options were to replace the name of the argument with _ in order to indicate that it's intentionally not used or to remove the argument altogether.  this was one of the few cases in which the argument wasn't actually used at all, so it could just be removed.  when it comes to the Turbine interface, there's inconsistent use of arguments, so it makes more sense to just unname those arguments.

jayjayjpg: I see and thank you for the outline on the linter upgrade leading to this change now and not previously, including this fix makes sense to me and thank you for the upgrade!

jayjayjpg (approved):
This reads great! Thank you for not only removing the `beta` tags, but also upgrading the linter in CI in one go! :sparkles:

PR #675, opened 2023-04-04
Validate environment exists before turbine app is deployed

### LINKED ISSUE #621: [Env: cli] Descriptive error when creating an app in a non existing env
## Context 

When creating an app in an environment that doesn't exist, we don't get a good error message.

```
meroxa apps deploy --env bad-aws
Validating your app.json...
	✔ Checked your language is "ruby"
	✔ Checked your application name is "test-rb"
Checking for uncommitted changes...
	✔ No uncommitted changes!
Error: could not find environment
```

## Acceptance

- [ ] Ensure API has a clearer message for when an app fails to deploy in an environment

lyuboxa:
## Description of change

Ensure environment exists when the `--env` flag is specified before the turbine deploy begins.

Fixes #621 

## Type of change

- [x]  New feature
- [ ]  Bug fix
- [x]  Refactor
- [ ]  Documentation

## How was this tested?

- [x]  Unit Tests
- [ ]  Tested in staging
- [ ]  Tested in minikube

## Demo

| before | after |
|--------|-------|
|||

### DIFF
--- a/cmd/meroxa/root/apps/deploy.go
+++ b/cmd/meroxa/root/apps/deploy.go
@@ -5,7 +5,7 @@ Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-	http://www.apache.org/licenses/LICENSE-2.0
+    http://www.apache.org/licenses/LICENSE-2.0
 
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
--- a/cmd/meroxa/root/apps/deploy.go
+++ b/cmd/meroxa/root/apps/deploy.go
@@ -13,6 +13,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
+
 package apps
 
 import (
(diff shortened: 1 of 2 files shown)


dianadoherty (approved):

cmd/meroxa/root/apps/deploy.go
```diff
@@ -5,15 +5,14 @@ Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-    http://www.apache.org/licenses/LICENSE-2.0
+	http://www.apache.org/licenses/LICENSE-2.0
```
janelletavares, line 8: intentional?

cmd/meroxa/root/apps/deploy.go
```diff
@@ -66,6 +66,22 @@ type deployApplicationClient interface {
...
+}
+
+func (e *environment) nameOrUUID() string {
+	switch {
+	case e.UUID != "":
+		return e.UUID
+	case e.Name != "":
+		return e.Name
+	default:
+		panic("bad state: name or uuid should be present")
```
janelletavares, line 81: i prefer not to panic because it feels like a bad customer experience if the CLI were to ever actually hit this

lyuboxa: I thought about this for a bit, the panic is a result of very unexpected state and we should be very vocal about it. Indeed this should never happen, but if it does.. it will blow up spectacularly, panic is used seldom, but it is used in the code base.

lyuboxa: Re bad customer experience, I'm not too worried, since this will grab the users attention and there is a higher chance of reporting it and engaging in a discussion.

At least this is my reasoning, let me know if you think its unreasonable.

raulb (approved):

cmd/meroxa/root/apps/deploy.go
janelletavares: Sorry, I missed your responses.  I don't think customers appreciate having to sort through stacktraces.  "Error: bad state" is as clear, imho.

lyuboxa: Yes, but it doesn't provide context why the state is bad. Ideally we want them to file a bug.

Issue #679, opened 2023-04-10
Allow a resource's token to be updated


janelletavares:
When it was added to the create command, it was not added to update at the same time, but it should function like any other field of credentials.

PR #680, opened 2023-04-10
Allow a resource's API token to be updated

janelletavares:
## Description of change

When the token flag was added to created, it should have been added to update at the same time

Fixes https://github.com/meroxa/cli/issues/679

## Type of change

- [ ]  New feature
- [x]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

## How was this tested?

- [x]  Unit Tests
- [ ]  Tested in staging
- [ ]  Tested in minikube

## Demo

| before | after |
|--------|-------|
|||


nassor (approved):
LGTM! I am just wondering why we are moving away from the ID(UUID).

jayjayjpg (approved):
This reads great, thank you for the super quick update!

janelletavares:
> LGTM! I am just wondering why we are moving away from the ID(UUID).

@nassor just to make it more readable on our end.  the customer never has ID, so the argument will only ever be name.  even though the Client function will accept ID, also.

Issue #687, opened 2023-04-14
[IR/Core Migration] Update turbine-go deploy implementation to use turbine-core


jmar910:
The `CreateDockerfile` implementation should be updated to:
+ [x] Call turbine-go to create the Dockerfile

The `SetupForDeploy` implementation should be updated to:
+ [x] Boot the `record` server
+ [x] Call turbine-go with the user's data app (See https://github.com/meroxa/turbine-go/issues/136)

The implementations that check the app (`GetResources`) should be updated to use app information from the `record` server

[EXAMPLE](https://github.com/meroxa/cli/blob/master/cmd/meroxa/turbine/ruby/deploy.go)

## Dependencies
+ [x] https://github.com/meroxa/turbine-go/issues/136

raulb:
Currently blocked on https://github.com/meroxa/turbine-go/issues/148.

PR #708, opened 2023-04-24
feat(apps): deploy for golang using turbine-core

raulb:
## Description of change

Fix https://github.com/meroxa/cli/issues/687

## Type of change

- [x]  New feature
- [ ]  Bug fix
- [x]  Refactor
- [ ]  Documentation

## How was this tested?

- [x]  Unit Tests
- [x]  Tested in staging
- [ ]  Tested in minikube

## Demo

**Before**

Apps deployment using core for golang wasn't supported.

**Now**

It is supported 😎 .

## Additional references

To test with your go application, you'd need to make sure:

1. Your demo go app uses the following imports (and latest of turbine-go):

```
	"github.com/meroxa/turbine-go/pkg/runner"
	"github.com/meroxa/turbine-go/pkg/turbine"
```

2. Update your `Process` function call to accept an error:

```
res, err := v.Process(rr, Anonymize{})
	if err != nil {
		return err
	}
```

## Demo

https://meroxa.slack.com/archives/C03F15EKFU3/p1682680029332049


cmd/meroxa/turbine/ruby/cli.go
```diff
@@ -3,41 +3,26 @@
...
 func New(l log.Logger, appPath string) turbine.CLI {
 	return &turbineRbCLI{
-		logger:            l,
-		appPath:           appPath,
-		runner:            server.NewRunServer(),
-		builder:           server.NewSpecBuilderServer(),
-		grpcListenAddress: server.ListenAddress,
+		logger:  l,
+		appPath: appPath,
+		turbineCore: turbine.Core{
```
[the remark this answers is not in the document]
raulb: Yeah. This is still a wip to get the commonality across languages. More refactors to come.

cmd/meroxa/turbine/golang/deploy.go
```diff
@@ -2,144 +2,119 @@ package turbinego
...
 }
 
 func (t *turbineGoCLI) CleanUpBuild(_ context.Context) {
-	utils.CleanupDockerfile(t.logger, t.appPath)
+	turbine.CleanupDockerfile(t.logger, t.appPath)
 }
 
-func (t *turbineGoCLI) SetupForDeploy(_ context.Context, _ string) (func(), error) {
-	return func() {}, nil
+func (t *turbineGoCLI) SetupForDeploy(ctx context.Context, gitSha, appName string) (func(), error) {
```
raulb, line 84: Still would like to refactor this method.

cmd/meroxa/root/apps/deploy.go
```diff
@@ -492,19 +492,13 @@ func (d *Deploy) deployApp(ctx context.Context, imageName, gitSha, specVersion s
...
-}
-
 // getAppImage will check what type of build needs to perform and ultimately will return the image name to use
 // when deploying.
 func (d *Deploy) getAppImage(ctx context.Context) (string, error) {
 	d.logger.StartSpinner("\t", "Checking if application has processes...")
 	var fqImageName string
 
-	needsToBuild, err := d.turbineCLI.NeedsToBuild(ctx, d.appName)
+	needsToBuild, err := d.turbineCLI.NeedsToBuild(ctx)
```
lyuboxa, line 501: So I don't think we will need this anymore, since we always have to build the app to generate the space. Fortunately, we don't need to compile it, but rather just run it with the spec builder server (similar to the run server) to get the spec.

cmd/meroxa/turbine/golang/deploy.go
lyuboxa: I think you can totally get away with a single command like this to get the spec https://github.com/meroxa/cli/pull/711/files#diff-b5881834513c853c914f1a532fccc86b12ed20be8eaff0c074f49570c1b33754R15-R23

lyuboxa: compared to build binary, run binary, cleanup binary

cmd/meroxa/root/apps/deploy.go
raulb: this is still necessary for other languages. For go, it checks if it has functions, through the spec, same as ruby. This is to see if it's needed to upload source etc...

lyuboxa: Ah yes, true, I think I confused this with another function

cmd/meroxa/root/apps/deploy.go
```diff
@@ -990,11 +980,13 @@ func (d *Deploy) Execute(ctx context.Context) error {
 	}
 
 	d.specVersion = d.flags.Spec
+
+	// Make the latest version default for Ruby (Go it's 0.2.0 ready, but not making it default for now)
 	if d.specVersion == "" && d.lang == ir.Ruby {
```
dianadoherty, line 985: I think we should do the same for go, because there's so many cli changes, that would not work without core

raulb: @dianadoherty yup! changed that!

cmd/meroxa/turbine/golang/deploy.go
```diff
@@ -2,144 +2,122 @@ package turbinego
...
 
-func (t *turbineGoCLI) SetupForDeploy(_ context.Context, _ string) (func(), error) {
-	return func() {}, nil
+func (t *turbineGoCLI) SetupForDeploy(ctx context.Context, gitSha string) (func(), error) {
+	source := rand.NewSource(time.Now().UnixNano())
+	rand := rand.New(source)
+	port := rand.Intn(65535-1024) + 1024 //nolint:gomnd
+	grpcListenAddress := fmt.Sprintf("localhost:%d", port)
+
+	go t.builder.RunWithAddress(ctx, grpcListenAddress)
```
lyuboxa, line 91: the grpc address can be set when the builder is instantiated

raulb: yes, you're right

raulb: @lyuboxa https://github.com/meroxa/cli/pull/708/commits/3502dd6a44da206109cf01d6869ced7e23f29fa9

dianadoherty (approved):

jmar910:
image

lyuboxa:
After this merge, we need to fix acceptance tests and rerun them to make sure this works.

PR #733, opened 2023-05-08
fix: upgrade turbine-core for latest py fix

jmar910:
## Description of change
Fixes python deploys on windows machines

## Type of change

- [ ]  New feature
- [x]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

## How was this tested?

- [ ]  Unit Tests
- [x]  Tested in staging
- [ ]  Tested in minikube


cmd/meroxa/root/apps/deploy.go
```diff
@@ -271,6 +271,9 @@ func createTarAndZipFile(src string, buf io.Writer) error {
 	tarWriter := tar.NewWriter(zipWriter)
 
 	err = filepath.Walk(appDir, func(file string, fi os.FileInfo, err error) error {
+		if err != nil {
+			return err
```
lyuboxa, line 275: returning this error will halt the tree walk, this is intended right?

jmar910: I figured it should be, otherwise we could get nil pointer errors when trying to operate on a file that isnt there.

PR #741, opened 2023-05-18
fix: init path on windows

jmar910:
## Description of change

Fixes https://github.com/meroxa/turbine-js/issues/305

Also cleans up upgrade/version commands in JS to rely on the newest version of turbine-js

## Type of change

- [ ]  New feature
- [x]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

## How was this tested?

- [ ]  Unit Tests
- [x]  Tested in staging
- [ ]  Tested in minikube

### DIFF
--- a/cmd/meroxa/turbine/javascript/version.go
+++ b/cmd/meroxa/turbine/javascript/version.go
@@ -4,22 +4,18 @@ import (
 	"context"
 	"fmt"
 
-	utils "github.com/meroxa/cli/cmd/meroxa/turbine"
+	"github.com/meroxa/cli/cmd/meroxa/turbine/javascript/internal"
 )
 
-// GetVersion calls turbine-js-cli to get the turbine-js-framework version used by the given app.
 func (t *turbineJsCLI) GetVersion(ctx context.Context) (string, error) {
-	stdOut, err := RunTurbineJS(ctx, t.logger, "version", t.appPath)
-	fmtErr := fmt.Errorf(
-		"unable to determine the version of turbine-js-framework used by the Meroxa Application at %s: %s",
-		t.appPath, err)
+	ver, err := internal.RunTurbineCmd(
+		ctx,
+		t.appPath,
+		internal.TurbineCommandVersion,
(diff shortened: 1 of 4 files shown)


cmd/meroxa/turbine/javascript/version.go
```diff
@@ -3,23 +3,27 @@ package turbinejs
...
-	if err != nil {
-		return "", fmtErr
-	}
+	cmd := internal.NewTurbineCmd(
+		ctx,
+		t.appPath,
+		internal.TurbineCommandVersion,
+		map[string]string{},
+	)
+	cmd.Stderr = nil
```
lyuboxa, line 18: this is not necessary, since you are using `CombinedOutput` which will replace the stderr/out with a buffer writer.

see https://cs.opensource.google/go/go/+/refs/tags/go1.20.4:src/os/exec/exec.go;l=1002-1004

lyuboxa: actually internal command sets it to the defaults. blah

maybe add this to internal from ruby to javascript https://github.com/meroxa/cli/blob/master/cmd/meroxa/turbine/ruby/internal/cmd.go#LL21C1-L27C2

cmd/meroxa/turbine/javascript/version.go
```diff
@@ -3,23 +3,27 @@ package turbinejs
...
 
-	version, err := utils.GetTurbineResponseFromOutput(stdOut)
+	stdout, err := cmd.CombinedOutput()
 	if err != nil {
-		err = fmtErr
+		return "", fmt.Errorf("cannot determine turbine-js node module version in %s", t.appPath)
 	}
-	return version, err
+
+	version := strings.TrimSpace(string(stdout))
```
lyuboxa, line 26: inline this in the return, e.g. `return strings.TrimSpace(..), nil`

lyuboxa (approved):

raulb (approved):
LGTM

PR #744, opened 2023-05-19
refactor: unify turbine-core functionality

raulb:
## Description of change

Fixes https://github.com/meroxa/product/issues/780

## Type of change

- [ ]  New feature
- [ ]  Bug fix
- [x]  Refactor
- [ ]  Documentation

## How was this tested?

- [x]  Unit Tests
- [x] Manual deployment
- [ ]  Tested in staging
- [ ]  Tested in minikube

## Additional references

This should be eventually moved to turbine-core. For now, modifying in the CLI as a first step.

### DIFF
--- a/cmd/meroxa/turbine/core.go
+++ b/cmd/meroxa/turbine/core.go
@@ -31,7 +31,7 @@ func NewCore() *Core {
 
 func (t *Core) Start(ctx context.Context) (string, error) {
 	var (
-		err error
+		err     error
 		retries = time.Tick(3 * time.Second)
 	)
 
@@ -39,7 +39,7 @@ func (t *Core) Start(ctx context.Context) (string, error) {
 
 	// NB: Spin until server is ready.
 	//     Ideally, server should communicate readyness but until then we wait.
-	for _ = range retries {
+	for range retries {
 		t.client, err = client.DialTimeout(t.grpcListenAddress, time.Second)
 		if err == nil {
 			return t.grpcListenAddress, nil
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -21,11 +21,6 @@ linters-settings:
       - hugeParam
   gocyclo:
     min-complexity: 15
-  gomnd:
-    settings:
-      mnd:
-        # don't include the "operation" and "assign"
-        checks: [argument,case,condition,return]
   govet:
     check-shadowing: true
     settings:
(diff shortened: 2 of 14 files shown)


cmd/meroxa/root/apps/deploy.go
```diff
@@ -396,11 +395,7 @@ func uploadFile(ctx context.Context, logger log.Logger, filePath, url string) er
 func (d *Deploy) createDeployment(ctx context.Context, imageName, gitSha, specVersion string) (*meroxa.Deployment, error) {
 	specStr, err := d.turbineCLI.GetDeploymentSpec(
```
lyuboxa, line 396: nit: might as well inline the args in the `(..)` rather than break the lines.

cmd/meroxa/turbine/core.go
```diff
@@ -0,0 +1,97 @@
+package turbine
```
lyuboxa, line 1: Do you want to put this in an internal package, so that its not exportable? 
e.g. `cmd/meroxa/turbine/internal`

cmd/meroxa/turbine/core.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+}
+
+func (t *Core) Start(ctx context.Context) string {
+	go t.builder.RunAddr(ctx, t.grpcListenAddress)
+	return t.grpcListenAddress
+}
+
+func (t *Core) Stop() (func(), error) {
+	c, err := client.DialTimeout(t.grpcListenAddress, time.Second)
```
lyuboxa, line 38: This feels like a side-effect of stopping, it's actually deceiving of what it does ;-)

lyuboxa: May be better to shove it in Start, but have to wait for the server to actually start.

cmd/meroxa/turbine/core.go
lyuboxa: Sounds good. Internal just mean it can’t be imported as a package outside of CLI, it is importable internally

cmd/meroxa/turbine/core.go
lyuboxa: @raulb i think there is a chance the client may try to connect before the go routine launches the server, otherwise the client may fail to connect.

PR #754, opened 2023-05-26
Command to delete flink jobs

janelletavares:
## Description of change

Don't let the papi db get cluttered without a way to clean it up. 

Fixes https://github.com/meroxa/turbine-project/issues/291

## Type of change

- [x]  New feature
- [ ]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

## How was this tested?

- [x]  Unit Tests
- [ ]  Tested in staging
- [ ]  Tested in minikube

## Demo

| before | after |
|--------|-------|
|||

### DIFF
--- a/cmd/meroxa/root/flink/remove_test.go
+++ b/cmd/meroxa/root/flink/remove_test.go
@@ -1,0 +1,107 @@
+/*
+Copyright © 2022 Meroxa Inc
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package flink
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"github.com/meroxa/meroxa-go/pkg/meroxa"
+	"reflect"
+	"testing"
+
+	"github.com/golang/mock/gomock"
+
+	"github.com/meroxa/cli/log"
+	"github.com/meroxa/cli/utils"
+	"github.com/meroxa/meroxa-go/pkg/mock"
+)
+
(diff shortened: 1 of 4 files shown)


justmisosoup:
@janelletavares How difficult possible would it be to rework the `meroxa flink` command to `meroxa jobs`?

ericcheatham:
@justmisosoup It's a fairly low effort change overall. Though that does raise a question about how we should refer to jobs in documentation/help text. Instead of 
> Remove flink job

should it be something like

> Remove job

that feels non-specific.

cc @janelletavares

justmisosoup:
@ericcheatham Good point. Articulating `Remove Flink job.` in the CLI documentation would be consistent with what we do with the `apps` commands. 

Today, that looks like:

```
Available Commands:
  deploy      Deploy a Turbine Data Application
  describe    Describe a Turbine Data Application
  init        Initialize a Turbine Data Application
  list        List Turbine Data Applications
  logs        View relevant logs to the state of the given Turbine Data Application
  ```

Looking forward to the future for when Flink is eventually encapsulated under `apps` we'll likely need to scrub Turbine Data Application from the CLI documentation and come up with something more generalised.

cc: @janelletavares

**Comment by janelletavares** [comment]
@justmisosoup I believe it's a one-liner to change `meroxa flink` to `meroxa jobs`. There are no generated docs for these commands yet because they are hidden, so nothing to update around there.

File: cmd/meroxa/root/flink/remove_test.go [hunk 1]
```diff
@@ -0,0 +1,107 @@
...
+			t.Fatalf("expected \"%s\" got \"%s\"", tt.err, err)
+		}
+
+		if tt.name != cc.args.NameOrUUID {
+			t.Fatalf("expected \"%s\" got \"%s\"", tt.name, cc.args.NameOrUUID)
+		}
+	}
+}
+
+func TestRemoveFlinkJobExecution(t *testing.T) {
```
jayjayjpg, line 59: Nice test coverage! :sparkles:

utils/tests.go
```diff
@@ -74,6 +74,13 @@ func GenerateResource() meroxa.Resource {
 	return GenerateResourceWithNameAndStatus("", "")
 }
 
+func GenerateFlinkJob() meroxa.FlinkJob {
```
jayjayjpg, line 77: I see, I was already wondering how you can mock out a flink job in those unit tests :eyes:

jayjayjpg (approved):
This reads great, thank you for putting this work together so quickly!

PR #772, opened 2023-06-09
Revert "Exit 1 when grpc requests fail"

janelletavares:
Reverts meroxa/cli#771


dianadoherty (approved):
Feel free to revert this, but I think there a lot of value to keeping it un-reverted. This PR helped uncover a really important bug that I think is worth leaving it around for :)

PR #780, opened 2023-06-17
Extract Connector IR from JAR

janelletavares:
## Description of change

Working with Flink Java to get an IR spec

Fixes https://github.com/meroxa/product/issues/854

## Type of change

- [x]  New feature
- [ ]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

## How was this tested?

- [x]  Unit Tests
- [x]  Tested in staging
- [ ]  Tested in minikube

## Demo

| before | after |
|--------|-------|
|||

### DIFF
--- a/cmd/meroxa/flink/deploy.go
+++ b/cmd/meroxa/flink/deploy.go
@@ -15,11 +15,13 @@ import (
 	"golang.org/x/mod/semver"
 )
 
-const irFilename = "meroxa-ir.json"
-const exactJavaVersion = "v11"
-const modeEnvVar = "MEROXA_PLATFORM"
-const outputEnvVar = "MEROXA_OUTPUT"
-const irVal = "EMIT_IR"
+const (
+	irFilename       = "meroxa-ir.json"
+	exactJavaVersion = "v11"
+	modeEnvVar       = "MEROXA_PLATFORM"
+	outputEnvVar     = "MEROXA_OUTPUT"
+	irVal            = "EMIT_IR"
+)
 
 func GetIRSpec(ctx context.Context, jarPath string, secrets map[string]string, l log.Logger) (*ir.DeploymentSpec, error) {
 	verifyJavaVersion(ctx, l)
--- a/cmd/meroxa/flink/deploy.go
+++ b/cmd/meroxa/flink/deploy.go
@@ -1,0 +1,133 @@
+package flink
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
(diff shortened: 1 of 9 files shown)


cmd/meroxa/flink/deploy.go
```diff
@@ -0,0 +1,96 @@
...
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		l.Warnf(ctx,
+			"warning: unable to verify local Java version is compatible with the Meroxa Platform; jar's must be compiled for %s",
+			exactJavaVersion)
+		return
+	}
+
+	// looks like 'openjdk version "11.0.19" 2023-04-18'
+	r := regexp.MustCompile(`version "([0-9.]+.[0-9.]+.[0-9.]+)"`)
```
samirketema, line 82: do we need to validate this _exact_ version? locally I have `openjdk64-11.0.18` installed, as well as the microsoft jdk 11

I wonder if we can generalize to just support java 11, though that is a bit more work

samirketema: Ah, I see that you're verifying `v11` in general. Though, it may not show in this exact format for the regex

samirketema: ok, I take this back, my local versions are matching this regex:
```
❯ java --version
openjdk 11.0.18 2023-01-17 LTS
OpenJDK Runtime Environment Microsoft-7208460 (build 11.0.18+10-LTS)
OpenJDK 64-Bit Server VM Microsoft-7208460 (build 11.0.18+10-LTS, mixed mode)
```

janelletavares: it's a bad constant name... I'll update to majorJavaVersion.

janelletavares: also, the output of -version and --version are slightly different :lolsob:

cmd/meroxa/flink/deploy.go
```diff
@@ -0,0 +1,96 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	_, err = os.Stat(irFilepath)
+	if err != nil {
+		// @TODO try the docker way because the jar is skinny
+		// Otherwise, there are no Meroxa* classes in this main class
+		return nil, nil
+	}
```
samirketema, line 53: nit: could be an inline err

samirketema (approved):
this is great. will take another look if you push up more changes

PR #786, opened 2023-06-27
Send valid IR spec with Flink deploy request

janelletavares:
## Description of change

Temporary workflow to validate how Flink Jobs and connectors can work together

Fixes https://github.com/meroxa/product/issues/860

## Type of change

- [x]  New feature
- [ ]  Bug fix
- [ ]  Refactor
- [ ]  Documentation

## How was this tested?

- [ ]  Unit Tests
- [ ]  Tested in staging
- [x]  Tested in minikube

## Demo

| before | after |
|--------|-------|
|||

### DIFF
--- a/cmd/meroxa/flink/deploy.go
+++ b/cmd/meroxa/flink/deploy.go
@@ -37,12 +37,12 @@ func GetIRSpec(ctx context.Context, jarPath string, secrets map[string]string, l
 		cmd.Environ(),
 		fmt.Sprintf("%s=%s", modeEnvVar, irVal),
 		fmt.Sprintf("%s=%s", outputEnvVar, irFilename))
-	_, err = cmd.CombinedOutput() // all java output goes to stderr, so that's fun
+	output, err := cmd.CombinedOutput() // all java output goes to stderr, so that's fun
 	defer func() {
 		_ = os.Remove(irFilepath)
 	}()
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("%s\n%v\n", output, err)
 	}
 
 	_, err = os.Stat(irFilepath)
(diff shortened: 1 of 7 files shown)


cmd/meroxa/flink/deploy.go
```diff
@@ -64,6 +69,31 @@ func GetIRSpec(ctx context.Context, jarPath string, secrets map[string]string, l
 	}
 
 	spec.Secrets = secrets
+	// workarounds to validate spec
```
ericcheatham, line 72: Created a task to follow up on this once the workarounds are no longer needed https://github.com/meroxa/product/issues/895

cmd/meroxa/flink/deploy.go
```diff
@@ -64,6 +69,31 @@ func GetIRSpec(ctx context.Context, jarPath string, secrets map[string]string, l
...
+	var sourceUUIDs []string
+	for _, cs := range spec.Connectors {
+		if cs.Type == ir.ConnectorDestination {
+			destinationUUID = cs.UUID
+		} else {
+			sourceUUIDs = append(sourceUUIDs, cs.UUID)
+		}
+	}
+
+	for _, u := range sourceUUIDs {
```
ericcheatham, line 88: This limits us to fanning in only, correct? I'm assuming that this will be revisited in a later epic.

cmd/meroxa/flink/deploy.go
janelletavares: thank you.  that should go under this epic https://github.com/meroxa/product/issues/868

cmd/meroxa/flink/deploy.go
janelletavares: oh yeah, definitely... this is a hack just to unblock us, but we don't actually have insight into the real DAG for any Flink Job at this point. every join and transformation would change which input eventually goes to which output. this is also a problem for when we try to figure out the output topics in the state machine in platform API.  if we have the dag, we can find the upstream entity, but without it, we're having to keep track of that relationship outside of the IR spec.

ericcheatham (approved):

PR #798, opened 2023-08-02
[Stateful Stream Processing] Check for the presence of meroxa resources 

ericcheatham:
## Description of change
Executing a `.jar` in order to extract IR information from without first checking if meroxa resources exist in the `.jar` has the side effect of running the job it its entirety if the job does not have meroxa resources. This may happen twice:
- Once during upload when inspecting
- Once after its been uploaded to our cluster and ran as would be expected

This could be a problem in out-of-scope execution of jobs may result in non-determinant results on the consumer's end

Fixes https://github.com/meroxa/product/issues/945

## Type of change

- [ ]  New feature
- [x]  Bug fix
- [x]  Refactor
- [ ]  Documentation

## How was this tested?

- [ ]  Unit Tests
- [x]  Tested in staging
- [ ]  Tested in minikube

### DIFF
--- a/cmd/meroxa/flink/deploy.go
+++ b/cmd/meroxa/flink/deploy.go
@@ -58,7 +58,6 @@ func GetIRSpec(ctx context.Context, jarPath string, secrets map[string]string, l
 	}
 
 	if usesMeroxa {
-
 		cwd, err := os.Getwd()
 		if err != nil {
 			return nil, err
--- a/cmd/meroxa/flink/deploy.go
+++ b/cmd/meroxa/flink/deploy.go
@@ -70,7 +70,8 @@ func GetIRSpec(ctx context.Context, jarPath string, secrets map[string]string, l
 	irFilepath := filepath.Join(cwd, irFilename)
 
 	// The submitted jar is executed with some special env vars set to inform the `MeroxaExecutionEnvironment` to
-	// short circuit execution and emit an IR spec instead https://github.com/meroxa/flink-platform-prototype/blob/main/src/main/java/com/meroxa/flink/MeroxaExecutionEnvironment.java#L64-L69
+	// short circuit execution and emit an IR spec instead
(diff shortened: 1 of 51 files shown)


janelletavares:
I don't think we need any of the doc updates right now, do we?

ericcheatham:
Probably not but also this happens every single time i run a make build on this repo.

samirketema:
@ericcheatham there's a env var `REBUILD_DOCS` in the makefile. It defaults to true, so that explains why it makes the changes when you do `make build`

ericcheatham:
@samirketema  
image


Good catch.

cmd/meroxa/flink/deploy.go
```diff
@@ -24,77 +26,105 @@ const (
 	irVal            = "EMIT_IR"
 )
 
+func doesJobUseMeroxaPlatform(ctx context.Context, jarPath string) (bool, error) {
```
samirketema, line 29: I think this approach is fine for now - do you want to leave a TODO comment (with a github issue) indicating that we need to verify that the `execute()` method is called on `MeroxaExecutionEnvironment`?

cmd/meroxa/flink/deploy.go
```diff
@@ -24,77 +26,105 @@ const (
...
-		return nil, fmt.Errorf("%s\n%v", output, err)
-	}
 
-	_, err = os.Stat(irFilepath)
-	if err != nil {
-		// @TODO try the docker way because the jar is skinny
-		// Otherwise, there are no Meroxa* classes in this main class
-		return nil, nil
-	}
+	if usesMeroxa {
```
samirketema, line 60: thoughts on checking if `!usesMeroxa` and then returning early? that way the rest of this code is not nested in a condition?

We could also add a comment explaining why we only do `java -jar` when it has the bindings

samirketema (commented):
looks good, some minor suggestions

cmd/meroxa/flink/deploy.go
ericcheatham: That's a good idea. Will definitely make it easier to read as well. 👏🏽

samirketema (approved):

PR #833, opened 2023-10-18
Mdpx switch apps commands

janelletavares:
## Description of change

Switching calls to new client calls

Fixes https://github.com/meroxa/mdpx/issues/583

## Type of change

- [ ]  New feature
- [ ]  Bug fix
- [x]  Refactor
- [ ]  Documentation

## How was this tested?

- [x]  Unit Tests
- [ ]  Tested in staging
- [ ]  Tested in minikube

## Demo

| before | after |
|--------|-------|
|||


utils/display/display.go
```diff
@@ -58,11 +64,13 @@ func PrintList(input interface{}, details Details) string {
 	for _, o := range objs {
 		bytes, err := json.Marshal(o)
 		if err != nil {
+			fmt.Printf("err1: %v\n", err)
```
nassor, line 67: what `err1` & `err2` stand for?

cmd/meroxa/root/apps/apps.go
```diff
@@ -30,6 +31,34 @@ import (
...
+
+const (
+	ApplicationStateInitialized ApplicationState = "initialized"
+	ApplicationStateDeploying   ApplicationState = "deploying"
+	ApplicationStatePending     ApplicationState = "pending"
+	ApplicationStateRunning     ApplicationState = "running"
+	ApplicationStateDegraded    ApplicationState = "degraded"
+	ApplicationStateFailed      ApplicationState = "failed"
+
+	// collectionName = "apps".
```
raulb, line 44: I assume this is not needed? Maybe we could remove in that case.

cmd/meroxa/root/apps/apps.go
```diff
@@ -30,6 +31,34 @@ import (
...
+//var displayDetails = display.Details{"Name": "name", "State": "state", .
+//"SpecVersion": "specVersion", "Created": "created", "Updated": "updated"}.
+
+// Application represents the Meroxa Application type within the Meroxa API.
+type Application struct {
+	ID          string           `json:"id"`
+	Name        string           `json:"name"`
+	State       ApplicationState `json:"state"`
+	Spec        string           `json:"spec"`
+	SpecVersion string           `json:"specVersion"`
```
raulb, line 56: I assume this is part of what we're already discussing in #mdpx, but I think we should continue using snakecase unless there's a very good reason to do so.

```suggestion
	SpecVersion string           `json:"spec_version"`
```

cmd/meroxa/root/apps/apps.go
```diff
@@ -58,11 +87,9 @@ func (*Apps) SubCommands() []*cobra.Command {
 		builder.BuildCobraCommand(&Describe{}),
 		builder.BuildCobraCommand(&Init{}),
 		builder.BuildCobraCommand(&List{}),
-		builder.BuildCobraCommand(&Logs{}),
```
raulb, line 61: I'd imagine the idea is to remove this for now until we have a better way to pull logs from mdpx?

raulb (approved):
@anna-cross looks good. I also wonder the same as @nassor here https://github.com/meroxa/cli/pull/833/files#r1368925587.

cmd/meroxa/root/apps/apps.go
anna-cross: This is not needed in this pr, i'll remove it for now!

utils/display/display.go
anna-cross: Good call out, I honestly dont remember why the errors are numbered....I'll remove the numbers

cmd/meroxa/root/apps/apps.go
anna-cross: Yup!!

raulb (approved):

PR #837, opened 2023-10-24
refactor: remove upgrade

raulb:
## Description of change

- Upgrade command was removed in another PR. This one removes its code.


anna-cross (approved):
Good catch!

PR #842, opened 2023-10-26
Mdpx remove meroxa go

anna-cross:
## Description of change

Removing dependencies on meroxa-go. There were a lot of unneeded old functions and processes that use meroxa-go that are no longer needed for mdpx. 

I've removed a lot of flink code but kept the files as stubs since we will still be implementing flink down the line 

Fixes https://github.com/meroxa/mdpx/issues/630

## Type of change

- [ ]  New feature
- [ ]  Bug fix
- [x]  Refactor
- [ ]  Documentation

## How was this tested?

- [x]  Unit Tests
- [ ]  Tested in staging
- [ ]  Tested in minikube


cmd/meroxa/global/global.go
```diff
@@ -52,10 +52,10 @@ const (
 	RefreshTokenEnv              = "REFRESH_TOKEN"
 	UserFeatureFlagsEnv          = "USER_FEATURE_FLAGS"
 	UserInfoUpdatedAtEnv         = "USER_INFO_UPDATED_AT"
-
-	TenantSubdomainEnv = "TENANT_SUBDOMAIN"
-	TenantEmailAddress = "TENANT_EMAIL_ADDRESS"
-	TenantPassword     = "TENANT_PASSWORD"
+	Actor                        = "ACTOR"
```
raulb, line 55: Where is this being used?

raulb (commented):
@anna-cross looks good. Aside from some linting issues, it should be ready to ship.

cmd/meroxa/global/global.go
anna-cross: Nope! Good catch, this can go!

